PR #17, opened 2022-05-25
Update go.mod to 1.16 to support go:embed

rmehta19:
While working on TLS Configuration Store library implementation, I ran into an error : 
"go:embed requires go1.16 or later (-lang was set to go1.14; check go.mod)".  Update go.mod to support go:embed.


matthewstevenson88 (approved):
Thanks for the PR @rmehta19!

PR #18, opened 2022-05-26
Static TLS Configuration Store Library Implementation

rmehta19:
This PR is a static implementation of the TLS Configuration Store Library. No calls to S2A are made. 

A note on tls_config_store_test.go
- The 'Equal' function cannot be used to validate the contents of rootCertPool because it is not yet part of a go release(https://cs.opensource.google/go/go/+/4aacb7ff0f103d95a724a91736823f44aa599634). A solution could be to temporarily copy the implementation of 'Equal' in the master branch of the golang. However the implementation accesses fields from the x509.CertPool struct, which are lowercase, hence private. Therefore, in our package we cannot implement Equal by defining an alias of CertPool, because CertPool fields are private.


internal/v2/tls_config_store/tls_config_store.go
```diff
@@ -0,0 +1,114 @@
...
+	clientCert []byte
+	//go:embed example_cert_key/server_cert.pem
+	serverCert []byte
+	//go:embed example_cert_key/client_key.pem
+	clientKey []byte
+	//go:embed example_cert_key/server_key.pem
+	serverKey []byte
+)
+
+func VerifyPeerCertificateFunc(instanceName string, pool *x509.CertPool) func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
```
rmehta19, line 25: I left the implementation here since the static implementation Certificate Verifier library is not implemented yet. In that library there will be two separate APIs(VerifyClientVertificateChain and VerifyServerCertificateChain)

internal/v2/tls_config_store/tls_config_store_test.go
```diff
@@ -0,0 +1,151 @@
...
+		// RootCAs 		    *x509.CertPool
+		InsecureSkipVerify          bool
+		CipherSuites	            []uint16
+		ClientSessionCache	    tls.ClientSessionCache
+		MinVersion	            uint16
+		MaxVersion		    uint16
+		CurvePreferences	    []tls.CurveID
+	}{
+		{
+			description: "static",
```
rmehta19, line 46: Would it be worthwhile to a nil test(Certificates == nil) to test this static implementation?

internal/v2/tls_config_store/example_cert_key/client_cert.pem
```diff
@@ -0,0 +1,25 @@
+-----BEGIN CERTIFICATE-----
```
rmehta19, line 1: I want to move this example_cert_key directory to be a subdirectory of internal/v2, since it will be used in testing of future libraries inside internal/v2. I was running into an error, similar to this discussion:https://github.com/golang/go/issues/46056. I wasn't able to resolve it during this PR, which is why I left example_cert_key inside the tls_config_store directory. Looking back, it may be with how I was formatting the string paths in the BUILD file. I'll try to move this folder one level up in the next PR.

rmehta19: From this link : https://go.googlesource.com/proposal/+/master/design/draft-embed.md#:~:text=An%20embed.Files,no%20embedded%20files, it seems to imply that data files cannot be in parent directories. I run into the following error when I put the data files in the interna/v2 directory:

"missing input file '//internal/v2/tls_config_store:s2a-go/internal/v2/example_cert_key/client_cert.pem' "

internal/v2/tls_config_store/tls_config_store_test.go
```diff
@@ -0,0 +1,151 @@
...
+			InsecureSkipVerify: true,
+			CipherSuites: nil,
+			ClientSessionCache: nil,
+			MinVersion: tls.VersionTLS13,
+			MaxVersion: tls.VersionTLS13,
+			CurvePreferences: nil,
+		},
+	} {
+		t.Run(tc.description, func(t *testing.T) {
+			config := GetTlsConfigurationForClient("abc")
```
rmehta19, line 58: "local_id" should be a field in the struct after this static implementation.

internal/v2/tls_config_store/example_cert_key/client_cert.pem
matthewstevenson88: Interesting. Let's leave it in whichever directory works for now, and add a TODO to come back to this later.

Rationale: At the end of the day when the full S2A-Go client is complete, I think these test files will only be used by the fake S2A implementation. So let's defer figuring this out exactly where to put these until we know exactly where they're needed.

internal/v2/tls_config_store/tls_config_store.go
```diff
@@ -0,0 +1,114 @@
+// Static implementation of TLS Configuration Store (no calls to S2Av2, Remote Signer Library, Certificate Verifier)
```
matthewstevenson88, line 1: nit: I think the package comment should reflect what the library does (as opposed to how it does it, e.g. by using static creds).

To that end, I would suggest saying something like "Builds TLS configurations that offload operations to S2Av2."

internal/v2/tls_config_store/tls_config_store.go
```diff
@@ -0,0 +1,114 @@
...
+package tls_config_store
```
matthewstevenson88, line 2: nit: package names in Go should be all one word I think, so this could be "tlsconfigstore"

internal/v2/tls_config_store/tls_config_store.go
matthewstevenson88: SGTM.

matthewstevenson88, line 25: nit: I think the VerifyPeerCertificateFunc should be unexported (i.e. "verify" should be lowercase) because we don't want to use it outside of this package.

Also, I think unexported functions should be moved to the bottom of the file (i.e. list the public ones first).

internal/v2/tls_config_store/tls_config_store.go
```diff
@@ -0,0 +1,114 @@
...
+		if cert.Subject.CommonName != instanceName {
+			return fmt.Errorf("certificate had Common Name %q, expected %q", cert.Subject.CommonName, instanceName)
+		}
+		return nil
+	}
+}
+
+
+func GetTlsConfigurationForClient(id string) tls.Config {
+	// Static implementation. TODO : Call S2Av2 for these values.
```
matthewstevenson88, line 53: Please structure your TODOs as follows:

// TODO(rmehta19): Call S2Av2 for these values.

Similarly throughout.

internal/v2/tls_config_store/tls_config_store.go
```diff
@@ -0,0 +1,114 @@
...
+
+		if cert.Subject.CommonName != instanceName {
+			return fmt.Errorf("certificate had Common Name %q, expected %q", cert.Subject.CommonName, instanceName)
+		}
+		return nil
+	}
+}
+
+
+func GetTlsConfigurationForClient(id string) tls.Config {
```
matthewstevenson88, line 52: Please add a top-level comment for each exported function.

For example:

matthewstevenson88, line 52: Why do we need the "id" parameter? Since it is unused at present should we remove it for now and can add in something later if need be?

internal/v2/tls_config_store/tls_config_store.go
```diff
@@ -0,0 +1,114 @@
...
+	cert, err := tls.X509KeyPair(clientCert, clientKey)
+	if err != nil {
+		log.Fatalf("Failed to get client cert")
+	}
+
+	rootCertPool := x509.NewCertPool()
+	rootCertPool.AppendCertsFromPEM(serverCert)
+
+	// Create mTLS credentials for client.
+	config := tls.Config {
```
matthewstevenson88, line 69: 2 micro-nits:

* Should we directly return rather than storing into the `config` variable?

* Should we return a pointer instead of the struct itself? In this case, we would need to write `return &tls.Config{...}`

internal/v2/tls_config_store/tls_config_store_test.go
```diff
@@ -0,0 +1,151 @@
...
+	//go:embed example_cert_key/server_cert.pem
+	serverCertpem []byte
+	//go:embed example_cert_key/client_key.pem
+	clientKeypem []byte
+	//go:embed example_cert_key/server_key.pem
+	serverKeypem []byte
+)
+
+
+// TODO(riyamehta) : In Client and Server test, verify contents of config.RootCAs once x509.CertPool.Equal function is officially released : https://cs.opensource.google/go/go/+/4aacb7ff0f103d95a724a91736823f44aa599634 .
```
matthewstevenson88, line 23: nit: Prefer using your Github username in the code here than your LDAP. :)

internal/v2/tls_config_store/tls_config_store_test.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+}
+
+func TestTLSConfigStoreServer(t *testing.T) {
+	// Setup for static server test.
+	cert, err := tls.X509KeyPair(serverCertpem, serverKeypem)
+	if err != nil {
+		t.Errorf("Test suite setup failed")
+	}
+	// certPool := x509.NewCertPool()
```
matthewstevenson88, line 95: Please remove commented out code.

internal/v2/tls_config_store/tls_config_store_test.go
```diff
@@ -0,0 +1,151 @@
...
+        // certPool.AppendCertsFromPEM(clientCertpem)
+
+	for _, tc := range []struct {
+		description		    string
+		Certificates                []tls.Certificate
+		ClientAuth		    tls.ClientAuthType
+		// ClientCAs		    *x509.CertPool
```
matthewstevenson88, line 102: Please remove commented out code.

internal/v2/tls_config_store/tls_config_store_test.go
```diff
@@ -0,0 +1,151 @@
...
+			MaxVersion: tls.VersionTLS13,
+			CurvePreferences: nil,
+		},
+	} {
+		t.Run(tc.description, func(t *testing.T) {
+			config := GetTlsConfigurationForClient("abc")
+			if got, want := config.Certificates[0].Certificate[0], tc.Certificates[0].Certificate[0]; !bytes.Equal(got, want) {
+				t.Errorf("config.Certificates[0].Certificate[0] = %v, want %v", got, want)
+			}
+			/*if got, want := config.RootCAs, tc.RootCAs; !got.Equal(want){
```
matthewstevenson88, line 62: Please remove commented out code.

internal/v2/tls_config_store/tls_config_store_test.go
matthewstevenson88: Per above comment, let's leave it out for now.

internal/v2/tls_config_store/tls_config_store.go
```diff
@@ -0,0 +1,114 @@
...
+		ClientSessionCache: nil,
+		MinVersion: min_version,
+		MaxVersion: max_version,
+		CurvePreferences: curve_preferences,
+	}
+	return config
+}
+
+
+func GetTlsConfigurationForServer(id string, server_name string) tls.Config {
```
matthewstevenson88, line 84: Same comments as for client-specific function.

internal/v2/tls_config_store/tls_config_store.go
```diff
@@ -0,0 +1,114 @@
...
+			return fmt.Errorf("certificate had Common Name %q, expected %q", cert.Subject.CommonName, instanceName)
+		}
+		return nil
+	}
+}
+
+
+func GetTlsConfigurationForClient(id string) tls.Config {
+	// Static implementation. TODO : Call S2Av2 for these values.
+	min_version := uint16(tls.VersionTLS13)
```
matthewstevenson88, line 54: For now, why don't we add these fields directly to the TLS config rather than having these temporary variables? It makes it a bit easier to read. :)

internal/v2/tls_config_store/tls_config_store_test.go
```diff
@@ -0,0 +1,151 @@
...
+
+// TODO(riyamehta) : In Client and Server test, verify contents of config.RootCAs once x509.CertPool.Equal function is officially released : https://cs.opensource.google/go/go/+/4aacb7ff0f103d95a724a91736823f44aa599634 .
+
+func TestTLSConfigStoreClient(t *testing.T) {
+	// Setup for static client test.
+	cert, err := tls.X509KeyPair(clientCertpem, clientKeypem)
+	if err != nil {
+		t.Errorf("Test suite setup failed")
+	}
+	// certPool := x509.NewCertPool()
```
matthewstevenson88, line 31: Please remove commented out code.

matthewstevenson88 (commented):
Thanks for the PR Riya! It's looking great! Left a handful of mostly-minor comments. Should be good to go after that.

internal/v2/tls_config_store/tls_config_store.go
rmehta19: Great! I'll add a TODO.

internal/v2/tls_config_store/example_cert_key/client_cert.pem
rmehta19: Will do, thanks for the rationale!

rmehta19:
> Thanks for the PR Riya! It's looking great! Left a handful of mostly-minor comments. Should be good to go after that.

Thanks for the review Matt!

internal/v2/tls_config_store/tls_config_store.go
rmehta19: Will do!

internal/v2/tls_config_store/tls_config_store.go
rmehta19: Great point, thanks!

internal/v2/tls_config_store/tls_config_store.go
rmehta19: Good point! I'll remove it for now.

internal/v2/tls_config_store/tls_config_store.go
```diff
@@ -0,0 +1,114 @@
...
+)
+
+func VerifyPeerCertificateFunc(instanceName string, pool *x509.CertPool) func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
+	return func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
+		if len(rawCerts) == 0 {
+			return fmt.Errorf("no certificate to verify")
+		}
+		cert, err := x509.ParseCertificate(rawCerts[0])
+		if err != nil {
+			return fmt.Errorf("x509.ParseCertificate(rawCerts[0]) returned error: %v", err)
```
tavishvaidya, line 32: Should we remove the variable name here and simply have the error stating that ParseCertificate failed? Adding the variable name makes it brittle for any future changes to the variable name.

internal/v2/tls_config_store/tls_config_store.go
```diff
@@ -0,0 +1,114 @@
...
+	// Static implementation. TODO : Call S2Av2 for these values.
+	min_version := uint16(tls.VersionTLS13)
+	max_version := uint16(tls.VersionTLS13)
+	var cipher_suites []uint16
+	var curve_preferences []tls.CurveID
+
+	// Static implementation. TODO : Call remote signer library for Private Key.
+	cert, err := tls.X509KeyPair(clientCert, clientKey)
+	if err != nil {
+		log.Fatalf("Failed to get client cert")
```
tavishvaidya, line 62: Should we "Failed to generate X509Keypair"? Get seems a bit overloaded since it can also mean that the function is trying to obtain a certificate from somewhere rather that simply parsing the provided one.

Also, while we are at it, let's log the "err" return by tls.X509KeyPair method.

tavishvaidya (commented):
Did a first pass, will do another one after all changes are in, thanks!

internal/v2/tls_config_store/tls_config_store.go
rmehta19: Thanks for pointing this out! I didn't know this was a feature but I found this(https://tip.golang.org/src/cmd/compile/internal/escape/escape.go), and if I interpreted it correctly, Go performs escape analysis, so in this case, if we return &tls.Config{...}, the Go compiler will allocate memory for it on the heap, whereas in other languages, returning a pointer to an object on the local stack would cause a memory error.

internal/v2/tls_config_store/tls_config_store.go
rmehta19: Will do!

internal/v2/tls_config_store/tls_config_store.go
rmehta19: Will do, thanks!

internal/v2/tls_config_store/tls_config_store.go
rmehta19: Will do, thanks!

rmehta19:
> Did a first pass, will do another one after all changes are in, thanks!
Thanks Tavish!

rmehta19:
Thanks for the comments! addressed them all in the latest push to this branch, let me know what you think!

internal/v2/tls_config_store/tls_config_store.go
```diff
@@ -0,0 +1,109 @@
...
+	//go:embed example_cert_key/client_key.pem
+	clientKey []byte
+	//go:embed example_cert_key/server_key.pem
+	serverKey []byte
+)
+
+// GetTlsConfigurationForClient returns a tls.Config instance for use by a client application.
+func GetTlsConfigurationForClient() *tls.Config {
+	// TODO(rmehta19): Call S2Av2 for config values.
+	var cipher_suites []uint16
```
matthewstevenson88, line 28: nit: can we get rid of the `cipher_suites` and `curve_preferences` variables for now since they are unused?

In the tls.Config, I think we can just omit those fields.

Similarly on the server side.

internal/v2/tls_config_store/tls_config_store.go
```diff
@@ -0,0 +1,109 @@
...
+		CurvePreferences: curve_preferences,
+	}
+}
+
+// GetTlsConfigurationForServer returns a tls.Config instance for use by a server application.
+func GetTlsConfigurationForServer() *tls.Config {
+	// TODO(rmehta19): Call S2Av2 for config values.
+	var cipher_suites []uint16
+	var curve_preferences []tls.CurveID
+	client_auth := tls.RequireAndVerifyClientCert
```
matthewstevenson88, line 59: Please move this directly into the tls.Config like above.

internal/v2/tls_config_store/tls_config_store_test.go
matthewstevenson88: Sorry I missed this - I think it's OK to omit this for now, since we'll be deferring this to S2A in the future.

internal/v2/tls_config_store/BUILD
```diff
@@ -0,0 +1,31 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")
+
+package(
+  default_visibility = ["//internal/v2:__subpackages__"],
+)
+
+#TODO(rmehta19) : Investigate how to use embedsrcs with data in a top level directory. If possible, move example_cert_key under s2a-go/internal/v2
```
matthewstevenson88, line 7: micro-nit: I think there should be a space between "#" and "TODO". In general, I suggest running "go mod tidy" on your code to clean up minor formatting stuff like this.

It's on my TODO list to add another CI that checks if there are formatting errors but I haven't had the time to do so yet. :(

matthewstevenson88 (approved):
LGTM with a few minor nits.

Thanks for the PR Riya!

internal/v2/tls_config_store/BUILD
rmehta19: Will do, thanks for the tip!

internal/v2/tls_config_store/tls_config_store_test.go
rmehta19: Sounds good, thanks!

PR #23, opened 2022-05-31
New example certificates with 20 year lifetime

rmehta19:


matthewstevenson88 (commented):
Thanks for the PR Riya! Should we also remove this TODO?

https://github.com/google/s2a-go/blob/8fbc40e6ec829d2be3ce0cd8f87649d2e7ecb99d/internal/v2/remote_signer/remote_signer_test.go#L10

PR #24, opened 2022-06-01
Transport Credentials Interface implementation

rmehta19:
Transport Credentials interface using static implementation of libraries.

### DIFF
--- a/internal/v2/s2av2.go
+++ b/internal/v2/s2av2.go
@@ -1,0 +1,101 @@
+// Package s2av2 provides the S2Av2 transport credentials used by a gRPC
+// application.
+package s2av2
+
+import (
+	"errors"
+	"context"
+	"net"
+	"crypto/tls"
+
+	"google.golang.org/grpc/credentials"
+	"github.com/google/s2a-go/internal/v2/tls_config_store"
+)
+
+const (
+	s2aSecurityProtocol = "s2av2"
+)
+
+type s2av2TransportCreds struct {
+	info     *credentials.ProtocolInfo
+	isClient bool
+	serverName string
+}
+
+// NewClientCreds returns a client-side transport credentials object that uses
+// the S2Av2 to establish a secure connection with a server.
+func NewClientCreds() (credentials.TransportCredentials, error) {
+	creds := &s2av2TransportCreds{
+		info: &credentials.ProtocolInfo{
+			SecurityProtocol: s2aSecurityProtocol,
+		},
+		isClient: true,
+		serverName: "",
+	}
+	return creds, nil
+}
+
+// NewServerCreds returns a server-side transport credentials object that uses
+// the S2Av2 to establish a secure connection with a client.
+func NewServerCreds() (credentials.TransportCredentials, error) {
+	creds := &s2av2TransportCreds{
+		info: &credentials.ProtocolInfo{
+			SecurityProtocol: s2aSecurityProtocol,
+		},
+		isClient: false,
+	}
+	return creds, nil
+}
+
+// ClientHandshake performs a client-side mTLS handshake using the S2Av2.
+func (c *s2av2TransportCreds) ClientHandshake(ctx context.Context, serverAuthority string, rawConn net.Conn) (net.Conn, credentials.AuthInfo, error) {
+	if !c.isClient {
+		return nil, nil, errors.New("client handshake called using server transport credentials")
+	}
+	// TODO(rmehta19): Remove the port from serverAuthority.
+	// TODO(rmehta19): Create a stub to S2Av2.
+	var config *tls.Config
+	if c.serverName == "" {
+		config = tlsconfigstore.GetTlsConfigurationForClient(serverAuthority)
+	} else {
+		config = tlsconfigstore.GetTlsConfigurationForClient(c.serverName)
+	}
+	creds := credentials.NewTLS(config)
+	return creds.ClientHandshake(context.Background(), serverAuthority, rawConn)
+}
+
+// ServerHandshake performs a server-side mTLS handshake using the S2Av2.
+func (c *s2av2TransportCreds) ServerHandshake(rawConn net.Conn) (net.Conn, credentials.AuthInfo, error) {
+	if c.isClient {
+		return nil, nil, errors.New("server handshake called using client transport credentials")
+	}
+	// TODO(rmehta19): Create a stub to S2Av2.
+	config := tlsconfigstore.GetTlsConfigurationForServer()
+	creds := credentials.NewTLS(config)
+	return creds.ServerHandshake(rawConn)
+}
+
+// Info returns protocol info of s2av2TransportCreds.
+func (c *s2av2TransportCreds) Info() credentials.ProtocolInfo {
+	return *c.info
+}
+
+// Clone makes a deep copy of s2av2TransportCreds.
+func (c * s2av2TransportCreds) Clone() credentials.TransportCredentials {
+	info := *c.info
+	serverName := c.serverName
+	return &s2av2TransportCreds{
+		info: &info,
+		isClient: c.isClient,
+		serverName: serverName,
+	}
+}
+
+// OverrideServerName sets the ServerName in the s2av2TransportCreds protocol
+// info. The ServerName MUST be a hostname.
+func (c *s2av2TransportCreds) OverrideServerName(serverNameOverride string) error {
+	// TODO(rmehta19): Remove the port from serverNameOverride.
+	c.info.ServerName = serverNameOverride
+	c.serverName = serverNameOverride
+	return nil
+}
--- a/internal/v2/s2av2_test.go
+++ b/internal/v2/s2av2_test.go
@@ -1,0 +1,156 @@
+package s2av2
+
+import (
+	"testing"
+	"context"
+	"github.com/google/go-cmp/cmp"
+	"google.golang.org/protobuf/testing/protocmp"
+)
+
+func TestNewClientCreds(t *testing.T) {
+	for _, tc := range []struct {
+		description string
+	}{
+		{
+			description: "static",
+		},
+	}{
+		t.Run(tc.description, func(t *testing.T){
+			c, err := NewClientCreds()
+			if err != nil {
+				t.Fatalf("NewClientCreds() failed: %v", err)
+			}
+			if got, want := c.Info().SecurityProtocol, s2aSecurityProtocol; got != want {
+				t.Errorf("c.Info().SecurityProtocol = %v, want %v", got, want)
+			}
+			_, ok := c.(*s2av2TransportCreds)
+			if !ok {
+				t.Fatal("the created creds is not of type s2av2TransportCreds")
+			}
+		})
+	}
+}
+
+func TestNewServerCreds(t *testing.T) {
+	for _, tc := range []struct {
+		description string
+	}{
+		{
+			description: "static",
+		},
+	}{
+		t.Run(tc.description, func(t *testing.T){
+			c, err := NewServerCreds()
+			if err != nil {
+				t.Fatalf("NewServerCreds() failed: %v", err)
+			}
+			if got, want := c.Info().SecurityProtocol, s2aSecurityProtocol; got != want {
+				t.Errorf("c.Info().SecurityProtocol = %v, want %v", got, want)
+			}
+			_, ok := c.(*s2av2TransportCreds)
+			if !ok {
+				t.Fatal("the created creds is not of type s2av2TransportCreds")
+			}
+		})
+	}
+}
+
+func TestClientHandshakeFail(t *testing.T) {
+	cc := &s2av2TransportCreds{isClient: false}
+	if _, _, err := cc.ClientHandshake(context.Background(), "", nil); err == nil {
+		t.Errorf("c.ClientHandshake(nil, \"\", nil) should fail with incorrect transport credentials")
+	}
+}
+
+func TestServerHandshakeFail(t *testing.T) {
+	sc := &s2av2TransportCreds{isClient: true}
+	if _, _, err := sc.ServerHandshake(nil); err == nil {
+		t.Errorf("c.ServerHandshake(nil) should fail with incorrect transport credentials")
+	}
+}
+
+func TestInfo(t *testing.T) {
+	c, err := NewClientCreds()
+	if err != nil {
+		t.Fatalf("NewClientCreds() failed: %v", err)
+	}
+	info := c.Info()
+	if got, want := info.SecurityProtocol, "s2av2"; got != want {
+		t.Errorf("info.SecurityProtocol=%v, want %v", got, want)
+	}
+}
+
+func TestCloneClient(t *testing.T) {
+	c, err := NewClientCreds()
+	if err != nil {
+		t.Fatalf("NewClientCreds() failed: %v", err)
+	}
+	cc := c.Clone()
+	s2av2Creds, ok := c.(*s2av2TransportCreds)
+	if !ok {
+		t.Fatal("the created creds is not of type s2av2TransportCreds")
+	}
+	s2av2CloneCreds, ok := cc.(*s2av2TransportCreds)
+	if !ok {
+		t.Fatal("the created clone creds is not of type s2aTransportCreds")
+	}
+	if got, want := cmp.Equal(s2av2Creds, s2av2CloneCreds, protocmp.Transform(), cmp.AllowUnexported(s2av2TransportCreds{})), true; got != want {
+		t.Errorf("cmp.Equal(%v, %v) = %v, want %v", s2av2Creds, s2av2CloneCreds, got, want)
+	}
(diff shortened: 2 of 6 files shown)


internal/v2/s2av2.go
```diff
@@ -0,0 +1,72 @@
...
+	return &s2av2TransportCreds{
+		info: &credentials.ProtocolInfo{
+			SecurityProtocol: s2aSecurityProtocol,
+		},
+		config: tlsconfigstore.GetTlsConfigurationForServer(),
+		isClient: false,
+	}
+}
+
+func (c *s2av2TransportCreds) ClientHandShake(ctx context.Context, serverAuthority string, rawConn net.Conn) (_ net.Conn, _ credentials.AuthInfo, err error) {
```
rmehta19, line 42: Since the original handshaker library offloaded the entire handshake to S2A, do we need to reimplement the handshaker interface to use it how it's used [here](https://github.com/google/s2a-go/blob/main/s2a.go#:~:text=secConn%2C%20authInfo%2C%20err%20%3A%3D%20chs.ClientHandshake(context.Background()))

internal/v2/s2av2.go
```diff
@@ -0,0 +1,72 @@
...
+	s2aSecurityProtocol = "s2av2"
+)
+
+type s2av2TransportCreds struct {
+	info     *credentials.ProtocolInfo
+	config   *tls.Config
+	isClient bool
+}
+
+func NewClientCreds() (*credentials.TransportCredentials) {
```
matthewstevenson88, line 22: nit: please add a comment for this function.

internal/v2/s2av2.go
```diff
@@ -0,0 +1,72 @@
...
+	return &s2av2TransportCreds{
+		info: &credentials.ProtocolInfo{
+			SecurityProtocol: s2aSecurityProtocol,
+		},
+		config: tlsconfigstore.GetTlsConfigurationForClient(),
+		isClient: true,
+	}
+}
+
+func NewServerCreds() *credentials.TransportCredentials {
```
matthewstevenson88, line 32: nit: please add a comment for this function.

internal/v2/s2av2.go
matthewstevenson88, line 22: Whoops you didn't do this below. :)

matthewstevenson88, line 22: nit: I would suggest removing the `*` in front of `credentials.TransportCredentials`. Generally you return an interface directly rather than a pointer to an interface (but you can return a pointer to an instance of a struct, e.g.).

matthewstevenson88, line 22: For future extensibility, I would return `(credentials.TransportCredentials, error)` instead even if we're not returning an error yet. This allows us to return errors in the future without having to change how people call this API.

internal/v2/s2av2.go
matthewstevenson88, line 42: The function name needs to exactly match the function name in the interface, so in this case it should be `ClientHandshake`.

internal/v2/s2av2.go
```diff
@@ -0,0 +1,72 @@
...
+	if !c.isClient {
+		return nil, nil, errors.New("client handshake called using server transport credentials")
+	}
+	// TODO(rmehta19): Create a stub to S2A.
+	// TODO(rmehta19): Offload verification of server cert to S2A.
+	// TODO(rmehta19): Offload signing the client's CertificateVerify message to S2A.
+}
+
+
+func (c *s2av2TransportCreds) ServerHandshake() () {
```
matthewstevenson88, line 52: The function signature needs to exactly match the function signature in the interface, so in this case it should be

```
ServerHandshake(rawConn net.Conn) (net.Conn, credentials.AuthInfo, error)
```

internal/v2/s2av2.go
matthewstevenson88: Nope, you won't have a handshaker interface for v2.

internal/v2/s2av2.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+func (c * s2av2TransportCreds) Clone() *credentials.TransportCredentials {
+	info := *c.info
+	return &s2aTransportCreds{
+		info: &info,
+		config: c.config.Clone()
+		isClient: c.isClient,
+	}
+}
```
matthewstevenson88, line 72: Don't forget about the `OverrideServerName` function as well:

https://github.com/google/s2a-go/blob/a941a3f6a219d2131d532a9abe27d7bd40244d35/s2a.go#L263

matthewstevenson88 (commented):
Thanks for the PR Riya! I left a few comments.

internal/v2/s2av2.go
rmehta19: I see, so instead of calling the handshaker interface methods like NewClientHandshaker and ClientHandshake, we make use of the libraries(tls_config_store, remote_signer and cert_verifier) which will make the RPCs to S2Av2, to just offload the certain operations in the handshake?

rmehta19:
Thanks for the comments Matt!

internal/v2/s2av2.go
matthewstevenson88: Roughly yes. In a bit more detail, the flow should be:

```
func ClientHandshake(...) ... {
    // 1. Create a grpc.ClientConn instance to the S2A, using the API in [A]
    // 2. Build a stub from the grpc.ClientConn (similar to what you did in the echo examples when you called grpcpb.EchoClient)
   // 3. Get a *tls.Config from the TLS config store. You will probably need to pass the stub to the TLS Config store's APIs.
   // 4. return credentials.NewTls(your TLS config).ClientHandshake(...)
}
```

and similarly for `ServerHandshake`.

[A] https://github.com/google/s2a-go/blob/a941a3f6a219d2131d532a9abe27d7bd40244d35/internal/handshaker/service/service.go#L41

matthewstevenson88: Disclaimer: this is just how I pictured it in my head, please push back if you prefer a different route :)

internal/v2/s2av2.go
```diff
@@ -0,0 +1,99 @@
...
+		info: &credentials.ProtocolInfo{
+			SecurityProtocol: s2aSecurityProtocol,
+		},
+		isClient: false,
+	}
+	return creds, nil
+}
+
+// ClientHandshake initiates a client-side mTLS handshake using the S2Av2.
+func (c *s2av2TransportCreds) ClientHandshake(ctx context.Context, serverAuthority string, rawConn net.Conn) (_ net.Conn, _ credentials.AuthInfo, err error) {
```
matthewstevenson88, line 48: Per offline discussion, we decided to test the success flows for `ClientHandshake` and `ServerHandshake` by locally starting up a client and server and having them talk to one another using the new transport credentials.

matthewstevenson88 (commented):
Thanks for the PR Riya! Generally looks good, just a handful of minor comments.

internal/v2/s2av2.go
rmehta19: Thanks for the explanation Matt!

internal/v2/s2av2.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/google/s2a-go/internal/v2/tls_config_store"
+)
+
+const (
+	s2aSecurityProtocol = "s2av2"
+)
+
+type s2av2TransportCreds struct {
+	info     *credentials.ProtocolInfo
+	config   *tls.Config
```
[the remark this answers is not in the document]
rmehta19: Thanks for pointing this out! Yes, definitely not needed, since we are creating a new config per handshake.

internal/v2/s2av2.go
rmehta19: Will do; I'll be sure to test in the next PR when we create an example go client / server using these s2av2transport credentials.

internal/v2/s2av2.go
```diff
@@ -0,0 +1,99 @@
...
+
+// ClientHandshake initiates a client-side mTLS handshake using the S2Av2.
+func (c *s2av2TransportCreds) ClientHandshake(ctx context.Context, serverAuthority string, rawConn net.Conn) (_ net.Conn, _ credentials.AuthInfo, err error) {
+	if !c.isClient {
+		return nil, nil, errors.New("client handshake called using server transport credentials")
+	}
+	c.config = tlsconfigstore.GetTlsConfigurationForClient()
+	// TODO(rmehta19): Create a stub to S2A.
+	// Offload verification of server cert to S2A.
+	c.config.VerifyPeerCertificate(c.config.RootCAs.Subjects(), nil)
```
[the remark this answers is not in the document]
matthewstevenson88: To clarify, in the long run, our code will not be explicitly calling VerifyPeerCertificate - we will just be passing the Golang equivalent of a std::function to the tls.Config and relying on the tls.Config to call this function at the right moment. Said differently, the structure will be something like this:

```
// Our custom verification function in the certverifier package.
func ourCustomVerifier(...) ... {
  // call S2Av2
}

func GetTlsConfigurationForClient() *tls.Config {
  return &tls.Config{
      ...
      VerifyPeerCertificate: ourCustomVerifier,
  }
}
```

Does that make sense?

Apologies if I misunderstood the question. :)

internal/v2/tls_config_store/tls_config_store.go
```diff
@@ -22,8 +21,13 @@ var (
 	serverKey []byte
 )
 
+func VerifyPeerCertificate(config *tls.Config) {
```
matthewstevenson88, line 24: This function should not be needed, I think we can remove it.

internal/v2/tls_config_store/tls_config_store.go
```diff
@@ -37,7 +41,7 @@ func GetTlsConfigurationForClient() *tls.Config {
 	// Create mTLS credentials for client.
 	return &tls.Config {
 		Certificates: []tls.Certificate{cert},
-		VerifyPeerCertificate: verifyPeerCertificateFunc("s2a_test_cert", rootCertPool), // TODO(rmehta19): Call cert verifier library.
+		VerifyPeerCertificate: certverifier.VerifyClientCertificateChain("s2a_test_cert", rootCertPool),
```
matthewstevenson88, line 44: When building the tls.Config for the client, we should use the VerifyServerCertificateChain function (because the client will be verifying the server's certificate).

This also means we need to pass in the hostname to that API. We can add a "serverHostname" string parameter to the GetTlsConfigurationForClient function. When we call GetTlsConfigurationForClient from ClientHandshake(), we can pass the serverAuthority parameter to it.

internal/v2/tls_config_store/tls_config_store.go
```diff
@@ -62,38 +67,11 @@ func GetTlsConfigurationForServer() *tls.Config {
 	// Create mTLS credentials for server.
 	return &tls.Config {
 		Certificates: []tls.Certificate{cert},
-		VerifyPeerCertificate: verifyPeerCertificateFunc("s2a_test_cert", certPool), // TODO(rmehta19): Call cert verifier library.
+		VerifyPeerCertificate: certverifier.VerifyServerCertificateChain("s2a_test_cert", "host", certPool),
```
matthewstevenson88, line 70: Similar to the above comment, this should be the VerifyClientCertificateChain API.

internal/v2/tls_config_store/tls_config_store.go
rmehta19: Thanks for catching this Matt, will do!

internal/v2/s2av2.go
```diff
@@ -0,0 +1,96 @@
+// Package s2av2 provides the S2Av2v2v2 transport credentials used by a gRPC
```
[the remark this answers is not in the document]
rmehta19: Thanks, will do!

internal/v2/s2av2.go
```diff
@@ -0,0 +1,88 @@
...
+	return creds, nil
+}
+
+// ClientHandshake initiates a client-side mTLS handshake using the S2Av2.
+func (c *s2av2TransportCreds) ClientHandshake(ctx context.Context, serverAuthority string, rawConn net.Conn) (net.Conn, credentials.AuthInfo, error) {
+	if !c.isClient {
+		return nil, nil, errors.New("client handshake called using server transport credentials")
+	}
+	// TODO(rmehta19): Create a stub to S2Av2.
+	config := tlsconfigstore.GetTlsConfigurationForClient("serverHostname")
```
matthewstevenson88, line 53: Sorry I missed this before - we have to be a little careful with how we populate the server hostname parameter (which we currently have hardcoded to "serverHostname").

I think we need to do the following:

1. Have a new "serverAuthority" variable in the `s2av2TransportCredentials` struct.
2. In `ClientHandshake`, use the following logic: if `c.serverAuthority` is empty, pass the `serverAuthority` parameter to `GetTlsConfigurationForClient`; otherwise, pass in `c.serverAuthority`.
3. In `OverrideServerName`, also populate `c.serverAuthority` with the overridden server name.

rmehta19:
Added serverName field to s2av2TransportCredentials and ServerName field to Client config obtained from GetTlsConfigurationForClient

internal/v2/s2av2.go
```diff
@@ -0,0 +1,99 @@
...
+
+type s2av2TransportCreds struct {
+	info     *credentials.ProtocolInfo
+	isClient bool
+	serverName string
+}
+
+// NewClientCreds returns a client-side transport credentials object that uses
+// the S2Av2 to establish a secure connection with a server.
+func NewClientCreds(serverName string) (credentials.TransportCredentials, error) {
```
matthewstevenson88, line 27: nit: I think we should omit the `serverName` parameter and leave it uninitialized (or explicitly initialize it to the empty string).

internal/v2/s2av2.go
```diff
@@ -0,0 +1,99 @@
...
+	serverName := c.serverName
+	return &s2av2TransportCreds{
+		info: &info,
+		isClient: c.isClient,
+		serverName: serverName,
+	}
+}
+
+// OverrideServerName sets the ServerName in the s2av2TransportCreds protocol
+// info.
```
matthewstevenson88, line 94: nit: Can you add a second sentence to this comment that says "The ServerName MUST be a hostname."?

internal/v2/s2av2.go
```diff
@@ -0,0 +1,99 @@
...
+
+// ClientHandshake performs a client-side mTLS handshake using the S2Av2.
+func (c *s2av2TransportCreds) ClientHandshake(ctx context.Context, serverAuthority string, rawConn net.Conn) (net.Conn, credentials.AuthInfo, error) {
+	if !c.isClient {
+		return nil, nil, errors.New("client handshake called using server transport credentials")
+	}
+	// TODO(rmehta19): Create a stub to S2Av2.
+	var config *tls.Config
+	if c.serverName == "" {
+		config = tlsconfigstore.GetTlsConfigurationForClient(serverAuthority)
```
rmehta19, line 58: Forgot about removing the port using net.SplithostPort, should that be done to serverAuthority here before passing to GetTlsConfigurationForClient?

internal/v2/s2av2_test.go
```diff
@@ -0,0 +1,146 @@
...
+	if err != nil {
+		t.Fatalf("NewClientCreds(\"\") failed: %v", err)
+	}
+	if got, want := c.Info().ServerName, ""; got != want {
+		t.Errorf("c.Info().ServerName = %v, want %v", got, want)
+	}
+	if err := c.OverrideServerName(wantServerName); err != nil {
+		t.Fatalf("c.OverrideServerName(%v) failed: %v", wantServerName, err)
+	}
+	if got, want := c.Info().ServerName, wantServerName; got != want {
```
matthewstevenson88, line 143: nit: Can we add one more check to ensure c.serverName is equal to wantServerName?

internal/v2/s2av2.go
matthewstevenson88: I don't want to block this PR any further so let's do this in a follow-up. Can we leave a TODO?

Note that we want to remove the port from `serverAuthority` and from `c.serverName`. We can maybe remove the port from `c.serverName` when it's set in OverrideServerName.

matthewstevenson88 (approved):
LGTM with a few nits. Thanks for the PR Riya!!

internal/v2/s2av2.go
rmehta19: Will do! So when a client wants to set the serverName, they will have to use OverrideServerName?

internal/v2/s2av2.go
rmehta19: Will do, thanks!

internal/v2/s2av2.go
matthewstevenson88: Correct.

internal/v2/s2av2.go
rmehta19: Added TODO to do this in follow up PR.

rmehta19:
Thanks for all the comments Matt!!

PR #26, opened 2022-06-03
Remove port from serverAuthority

rmehta19:


tavishvaidya (approved):

internal/v2/s2av2.go
```diff
@@ -94,8 +98,12 @@ func (c * s2av2TransportCreds) Clone() credentials.TransportCredentials {
 // OverrideServerName sets the ServerName in the s2av2TransportCreds protocol
 // info. The ServerName MUST be a hostname.
 func (c *s2av2TransportCreds) OverrideServerName(serverNameOverride string) error {
-	// TODO(rmehta19): Remove the port from serverNameOverride.
-	c.info.ServerName = serverNameOverride
-	c.serverName = serverNameOverride
+	// Remove the port from serverNameOverride.
+	serverName, _, err := net.SplitHostPort(serverNameOverride)
```
matthewstevenson88, line 102: If we don't already have them, can we add following test cases?

1. Calling OverrideServerName with an empty string.
2. Calling OverrideServerName with a non-empty string that makes net.SplitHostPort return an error.
3. Calling OverrideServerName with a non-empty string that has both a host and a port and net.SplitHostPort returns just the host.
4. Calling OverrideServerName with a non-empty string that has just a host and net.SplitHostPort returns just the host.

To do this, it might make sense to change the existing test into one of the "for loop style" tests.

rmehta19: Will do!

matthewstevenson88 (approved):

PR #27, opened 2022-06-03
[WIP] go echo client / server using s2av2 Transport Credentials

rmehta19:


rmehta19:
I started up the client and server in two terminals, here is the output:

bazel run client:
Screenshot 2022-06-03 3 02 59 PM

bazel run server:
Screenshot 2022-06-03 3 03 45 PM

rmehta19:
In the last few pushes the client / server continued to work. The Bazel tests were failing because I had changed the internal/v2 BUILD file in my local workspace, but missed adding that to the commit, once added, those tests starting passing here. Currently, the Golang tests are failing here because it is not able to find the packages:
image

example/BUILD
```diff
@@ -0,0 +1,30 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_library")
+
+go_binary(
+  name = "client",
+  srcs = ["client.go"],
+  deps = [
+    "//internal/v2:s2av2",
```
tavishvaidya, line 7: Should there be a dependency on ":echo" target as well?

rmehta19: Thanks for catching this! I'm not sure it needs to be here, because the echo package has the libraries to run the server. Related to the "go build" error I'm having, to test this, I added the dependency in my workspace. It didn't change the error I get when I run go build. It also didn't affect the bazel run client / server (continues to work correctly)

rmehta19:
Change package names to match directory names(@matthewstevenson88). Also, I moved client and server to their own sub-directories to avoid re-declaration of main error raised by go compiler. Confirmed that client and server still work:

bazel run server:
Screenshot 2022-06-06 9 49 42 AM

bazel run client:
Screenshot 2022-06-06 9 50 09 AM

example/README.md
```diff
@@ -0,0 +1,2 @@
+This example implements a gRPC Go Echo client and server which use self - signed X.509 certificates and s2av2TransportCreds(implementation of
```
matthewstevenson88, line 1: 2 micro-nits:

1. Please remove spaces between self and signed, i.e. it should all be one word "self-signed".
2. Please leave a space between s2av2TransportCreds and the start of the parenthesis.

example/client/BUILD
```diff
@@ -0,0 +1,11 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_binary")
+
+go_binary(
+  name = "client",
+  srcs = ["client.go"],
```
matthewstevenson88, line 5: Can we mark this BUILD target as "testonly = True"?

Similarly for the server-side API and for echo.go.

example/client/client.go
```diff
@@ -0,0 +1,50 @@
...
+	conn, err := grpc.Dial(*serverAddr, opts...)
+	if err != nil {
+		log.Fatalf("Client: failed to connect: %v", err)
+	}
+	defer conn.Close()
+	c := pb.NewEchoClient(conn)
+	log.Printf("Client: connected to: %s", *serverAddr)
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTimeout)
+	defer cancel()
+	msg := "Hello!"
```
matthewstevenson88, line 40: Can we make the message something that definitely would not already appear in logs? For example, "Hello S2Av2 user!"?

example/client/client.go
```diff
@@ -0,0 +1,50 @@
...
+	r, err := c.Echo(ctx, &pb.EchoRequest{Msg: msg})
+	if err != nil {
+		log.Fatalf("Client: failed to send echo message: %v", err)
+	}
+	log.Printf("Client: recieved message from server: %s", r.GetMsg())
```
matthewstevenson88, line 45: micro-nit: typo in "received"

example/echo/echo.go
```diff
@@ -0,0 +1,17 @@
...
+
+	pb "github.com/google/s2a-go/example/proto/echo_go_proto"
+)
+
+// Server is an echo server used for testing.
+type Server struct {
+	pb.UnimplementedEchoServer
+}
+
+func (s *Server) Echo(ctx context.Context, req *pb.EchoRequest) (*pb.EchoResponse, error) {
```
matthewstevenson88, line 15: Please add a top-level comment for this function.

example/proto/BUILD
```diff
@@ -0,0 +1,25 @@
...
+)
+
+# Generates Go code of proto messages.
+go_proto_library(
+  name = "echo_go_proto",
+  importpath = "github.com/google/s2a-go/example/proto/echo_go_proto",
+  protos = [":echo_proto"],
+)
+
+# Generates gRPC services
```
matthewstevenson88, line 20: micro-nit: please add a period at the end of this sentence.

example/server/server.go
```diff
@@ -0,0 +1,34 @@
...
+
+func runServer(listenPort *string) {
+	creds, err := v2.NewServerCreds()
+	listener, err := net.Listen("tcp", *listenPort)
+	if err != nil {
+		log.Fatalf("failed to listen on addres %s: %v", *listenPort, err)
+	}
+	s := grpc.NewServer(grpc.Creds(creds))
+	log.Printf("Server: starting gRPC Echo Server at: %s", *listenPort)
+	log.Printf("Server: started listening on: %s", *listenPort)
```
matthewstevenson88, line 25: Should we omit one of these 2 print statements, since they're essentially doing the same thing?

matthewstevenson88 (commented):
Thanks for the PR Riya! It looks good in general, just a few minor comments.

matthewstevenson88 (approved):

PR #28, opened 2022-06-03
Regenerate echo example proto files

rmehta19:
TODO: echo is a shell command, and also part of the filename, need to find way to escape it.


rmehta19:
"echo" begin a shell command doesn't cause problems with running these commands; the editor highlights the word echo as if it is a shell command, but they run just fine

tavishvaidya (approved):

matthewstevenson88 (approved):

PR #30, opened 2022-06-07
Fake s2av2 service layer

rmehta19:
Fake SetupSession, GetTlsConfiguration


rmehta19:
Fake Service running using bazel run:
image

rmehta19:
This PR implements the SetUpSession RPC and also GetTlsConfiguration(called from switch in SetUpSession). The remaining 3 RPCs to be faked are commented in the switch as well. There is also a fake s2av2 service implementation.

internal/v2/fake_s2av2/fake_s2av2_test.go
```diff
@@ -0,0 +1,113 @@
+package fake_s2av2
+
+import (
+	"testing"
+	s2av2pb "github.com/google/s2a-go/internal/proto/v2/s2a_go_proto"
+	commonpb "github.com/google/s2a-go/internal/proto/v2/common_go_proto"
+)
+
```
rmehta19, line 8: Should we have this test suite actually setup, run and tear down the service? Is there a preferred Go package for this? I found the net/http package, it looks like it's for testing web applications, could we use it here?

internal/v2/fake_s2av2/fake_s2av2.go
```diff
@@ -0,0 +1,105 @@
...
+					TlsConfiguration: &s2av2pb.GetTlsConfigurationResp_ServerTlsConfiguration_{
+						&s2av2pb.GetTlsConfigurationResp_ServerTlsConfiguration{
+							CertificateChain: []string{
+							string(serverCert),
+							},
+							MinTlsVersion: commonpb.TLSVersion_TLS_VERSION_1_3,
+							MaxTlsVersion: commonpb.TLSVersion_TLS_VERSION_1_3,
+							HandshakeCiphersuites: []commonpb.HandshakeCiphersuite{},
+							RecordCiphersuites: []commonpb.RecordCiphersuite{},
+							TlsResumptionEnabled: false,
```
matthewstevenson88, line 94: Good catch!

internal/v2/fake_s2av2_service/fake_s2av2_service.go
```diff
@@ -0,0 +1,31 @@
...
+	"net"
+	"flag"
+	"log"
+	"google.golang.org/grpc"
+	"github.com/google/s2a-go/internal/v2/fake_s2av2"
+	s2av2pb "github.com/google/s2a-go/internal/proto/v2/s2a_go_proto"
+)
+
+var (
+	listenPort = flag.String("listen_port", ":8080", "Fake S2Av2 service address port.")
```
matthewstevenson88, line 13: nit: s/service/server

For consistency throughout the file.

matthewstevenson88, line 13: nit: Can we just call the variable and flag "port" instead of "listenPort"? The former is more common.

internal/v2/fake_s2av2/fake_s2av2_test.go
matthewstevenson88: Ya I think we should change up how the unit tests are structure here (sorry in advance if this is painful to refactor - it will be annoying in this PR but the follow-ups for the other RPCs will be easy because we'll do the setup here).

Step #1: In the test file, create the following function:
```
func startFakeS2Av2Server() (address string, stop func(), err error) {
  // Pick an unused port, set address to "localhost:" + that port
  s := grpc.NewServer()
  // Register your fake s2av2 implementation with this server
  go func() {
     if err := s.Serve(listener); err != nil {
        // log error
      }
   }
  return address, func() { s.Stop() }, nil
}
``

Step #2: Build a table test with 3 fields in the struct: description, request, expectedResponse. The body of the test would go something like this:

```
address, stop, err := startFakeS2Av2Server()
// Might need to use a wait group or similar to make sure there is no flakiness
// if the server is slop to start up

// Create a stream to the server, send the request, get the response, check that the response is the expected one, and close the stream

stop()
```

Step #3: How to efficiently compare protos (apologies I didn't think to mention this yesterday when we were talking): we should use the `cmp.Diff` and `protocmp.Transform()` APIs, e.g. as is done here:

https://github.com/google/s2a-go/blob/e978c1436cecf397b4c1178c7e56519941b7d196/internal/handshaker/handshaker_test.go#L1138

Happy to have a meeting to discuss any of this since the details are a bit complicated. :)

**Review by matthewstevenson88 — COMMENTED** [review]
Thanks for the PR Riya!

File: internal/v2/fake_s2av2/BUILD [hunk 4]
```diff
@@ -0,0 +1,29 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")
+
+package(
+  default_visibility = ["//internal/v2:__subpackages__"],
+)
+
+go_library(
+  name = "fake_s2av2",
+  srcs = ["fake_s2av2.go"],
```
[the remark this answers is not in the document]
rmehta19: Will do! I'll add a todo to go through the previous code I've submitted and fix this if I find any

internal/v2/fake_s2av2/fake_s2av2.go
```diff
@@ -0,0 +1,105 @@
...
+		case *s2av2pb.SessionReq_OffloadPrivateKeyOperationReq:
+			// TODO(rmehta19): Implement fake.
+		case *s2av2pb.SessionReq_OffloadResumptionKeyOperationReq:
+			// TODO(rmehta19): Implement fake.
+		case *s2av2pb.SessionReq_ValidatePeerCertificateChainReq:
+			// TODO(rmehta19): Implement fake.
+		default:
+			return fmt.Errorf("SessionReq.ReqOneof has unexpected type %T", x)
+		}
+		err1 := stream.Send(s)
```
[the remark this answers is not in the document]
rmehta19: Will do, thanks!

internal/v2/fake_s2av2/fake_s2av2.go
```diff
@@ -0,0 +1,105 @@
...
+		if err1 != nil {
+			log.Fatalf("Fake S2A Service: failed to send SessionResp: %v", err1)
+			return err1
```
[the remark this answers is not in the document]
rmehta19: Thanks for pointing this out, I didn't know there was a difference between the two! I changed all the logging in SetUpSession to use Errorf, so that the error will be returned

internal/v2/fake_s2av2/fake_s2av2.go
rmehta19: Thanks!

internal/v2/fake_s2av2/fake_s2av2.go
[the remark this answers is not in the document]
rmehta19: We'll use log.Printf as discussed offline

internal/v2/fake_s2av2_service/BUILD
```diff
@@ -0,0 +1,12 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_binary")
+
+go_binary(
+  name = "fake_s2av2_service",
+  srcs = ["fake_s2av2_service.go"],
```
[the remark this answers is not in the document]
rmehta19: Will do! I'll rename the folder so we don't run into the package name directory name mismatch issue

internal/v2/fake_s2av2_service/fake_s2av2_service.go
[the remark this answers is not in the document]
rmehta19: Will do, I changed the port to :0, which I believe will choose a free, available port.

internal/v2/fake_s2av2/fake_s2av2_test.go
rmehta19: Thanks so much for this explanation!! It was super helpful :)

internal/v2/fakes2av2/fakes2av2.go
```diff
@@ -0,0 +1,115 @@
...
+	commonpb "github.com/google/s2a-go/internal/proto/v2/common_go_proto"
+)
+var (
+	//go:embed example_cert/client_cert.pem
+	clientCert []byte
+	//go:embed example_cert/server_cert.pem
+	serverCert []byte
+)
+
+// Server is a Fake S2A Server for testing.
```
matthewstevenson88, line 17: micro-nit: s/Fake/fake

internal/v2/fakes2av2/fakes2av2.go
```diff
@@ -0,0 +1,115 @@
...
+	for {
+		req, err := stream.Recv()
+		if err != nil {
+			log.Printf("Fake S2A Service: failed to receive SessionReq: %v", err)
+			return err
+		}
+		// Call one of the 4 possible RespOneof's
+		// TODO(rmehta19): Consider validating the body of the request.
+		var resp *s2av2pb.SessionResp
+		var err0 error
```
matthewstevenson88, line 34: nit: we don't need to declare err0 here, we can just redefine the error each time we get one

internal/v2/fakes2av2/fakes2av2.go
```diff
@@ -0,0 +1,115 @@
...
+		switch x := req.ReqOneof.(type) {
+		case *s2av2pb.SessionReq_GetTlsConfigurationReq:
+			if resp, err0 = getTlsConfiguration(req.GetGetTlsConfigurationReq()); err0 != nil {
```
matthewstevenson88, line 37: micro-nit: s/err0/err

It's common practice in Go to overwrite the same error variable over and over again.

internal/v2/fakes2av2/fakes2av2_test.go
```diff
@@ -0,0 +1,177 @@
...
+	"github.com/google/go-cmp/cmp"
+	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/protobuf/testing/protocmp"
+
+	s2av2pb "github.com/google/s2a-go/internal/proto/v2/s2a_go_proto"
+	commonpb "github.com/google/s2a-go/internal/proto/v2/common_go_proto"
+)
+
+const (
+	defaultTimeout = 30.0 * time.Second
```
matthewstevenson88, line 20: nit: maybe decrease to 10?

internal/v2/fakes2av2/fakes2av2_test.go
```diff
@@ -0,0 +1,177 @@
...
+
+func startFakeS2Av2Server(wg *sync.WaitGroup) (address string, stop func(), err error) {
+	// Pick unused port.
+	listener, err := net.Listen("tcp", ":0")
+	address = listener.Addr().String()
+	if err != nil {
+		log.Fatalf("failed to listen on port %s: %v", listener.Addr().String(), err)
+	}
+	s := grpc.NewServer()
+	log.Printf("Server: started gRPC Fake S2Av2 Server at port: %s", listener.Addr().String())
```
matthewstevenson88, line 31: micro-nit: s/at port/on address

since listener.Addr().String() is the full address

Similarly below.

internal/v2/fakes2av2/fakes2av2_test.go
```diff
@@ -0,0 +1,177 @@
...
+			},
+		},
+	}{
+		t.Run(tc.description, func(t *testing.T) {
+			// Start up Fake S2Av2 Server.
+			var wg sync.WaitGroup
+			wg.Add(1)
+			address, stop, err := startFakeS2Av2Server(&wg)
+			wg.Wait()
+			// Create stream to server.
```
matthewstevenson88, line 136: nit: Please add a blank line between the `wg.Wait()` line and the comment line, just to break up the text a little bit. Similarly below.

internal/v2/fakes2av2/fakes2av2_test.go
```diff
@@ -0,0 +1,177 @@
...
+			if err != nil {
+				t.Fatalf("Client: failed to connect: %v", err)
+			}
+			defer conn.Close()
+			c := s2av2pb.NewS2AServiceClient(conn)
+			log.Printf("Client: connected to: %s", address)
+			ctx, cancel := context.WithTimeout(context.Background(), defaultTimeout)
+			defer cancel()
+			// Setup bidrectional streaming session.
+			s2av2_opts := []grpc.CallOption{}
```
matthewstevenson88, line 152: nit: can we call this callOpts instead? Generally, underscores should be avoided in variable names.

internal/v2/fakes2av2/fakes2av2_test.go
```diff
@@ -0,0 +1,177 @@
...
+			cstream, err0 := c.SetUpSession(ctx, s2av2_opts...)
```
matthewstevenson88, line 153: micro-nit: please use `err` instead of `err0`

internal/v2/fakes2av2/fakes2av2_test.go
```diff
@@ -0,0 +1,177 @@
...
+				t.Fatalf("Client: failed to setup bidirectional streaming RPC session: %v", err0)
+			}
+			log.Printf("Client: set up bidirectional streaming RPC session.")
+			// Send request.
+			if err := cstream.Send(tc.request); err != nil {
+				t.Fatalf("Client: failed to send SessionReq: %v", err)
+			}
+			log.Printf("Client: sent SessionReq")
+			// Get the Response.
+			resp, err1 := cstream.Recv()
```
matthewstevenson88, line 164: micro-nit: please use `err` instead of `err1`

internal/v2/fakes2av2/fakes2av2_test.go
```diff
@@ -0,0 +1,177 @@
...
+			if err1 != nil {
+				t.Fatalf("Client: failed to receive SessionResp: %v", err1)
+			}
+			log.Printf("Client: recieved SessionResp")
+			if diff := cmp.Diff(resp, tc.expectedResponse, protocmp.Transform()); diff != "" {
+				t.Errorf("cstream.Recv() returned incorrect SessionResp, (-want +got):\n%s", diff)
+			} else {
```
matthewstevenson88, line 171: I think it's OK to skip the else block - this print statement doesn't tell us anything because we already know if the test failed or not by the time we get here.

matthewstevenson88 (commented):
Thanks Riya! The PR generally looks good, just a handful of minor comments. :)

matthewstevenson88 (approved):

PR #31, opened 2022-06-08
TLS Config Store makes calls to Fake S2A Service

rmehta19:
In this PR:
-TLS Config Store modified to call S2Av2 for Client TLS Configuration and Server TLS Configuration. 
-Test suite modified to start / stop fake S2Av2 Server on port 8008.
-Cleaned up some error messages to be more clear.
-Removed fakes2av2_server binary(accidentally added through running go build in previous PR)

### DIFF
--- a/internal/v2/tls_config_store/tls_config_store.go
+++ b/internal/v2/tls_config_store/tls_config_store.go
@@ -55,7 +55,6 @@ func GetTlsConfigurationForClient(serverHostname string, cstream s2av2pb.S2AServ
 	tlsConfig := resp.GetGetTlsConfigurationResp().GetClientTlsConfiguration()
 
 	var cert tls.Certificate
-	lenChain := len(tlsConfig.CertificateChain)
 	for i, v := range tlsConfig.CertificateChain {
 		// Populate Certificates field
 		block, _ := pem.Decode([]byte(v))
@@ -67,7 +66,7 @@ func GetTlsConfigurationForClient(serverHostname string, cstream s2av2pb.S2AServ
 			return nil, err
 		}
 		cert.Certificate = append(cert.Certificate, x509Cert.Raw)
-		if i == lenChain {
+		if i == 0 {
 			cert.Leaf = x509Cert
 		}
 	}
@@ -135,7 +134,6 @@ func GetTlsConfigurationForServer(cstream s2av2pb.S2AService_SetUpSessionClient)
 	tlsConfig := resp.GetGetTlsConfigurationResp().GetServerTlsConfiguration()
 
 	var cert tls.Certificate
-	lenChain := len(tlsConfig.CertificateChain)
 	for i, v := range tlsConfig.CertificateChain {
 		// Populate Certificates field
 		block, _ := pem.Decode([]byte(v))
@@ -147,7 +145,7 @@ func GetTlsConfigurationForServer(cstream s2av2pb.S2AService_SetUpSessionClient)
 			return nil, err
 		}
 		cert.Certificate = append(cert.Certificate, x509Cert.Raw)
-		if i == lenChain {
+		if i == 0 {
 			cert.Leaf = x509Cert
 		}
 	}
@@ -183,6 +181,7 @@ func GetTlsConfigurationForServer(cstream s2av2pb.S2AService_SetUpSessionClient)
 	}, nil
 }
 
+// TODO(rmehta19): Test every possible case of input/outputs for this function.
 func getTLSMinMaxVersionsClient(tlsConfig *s2av2pb.GetTlsConfigurationResp_ClientTlsConfiguration) (uint16, uint16, error) {
 	// Map S2Av2 TLSVersion to consts defined in tls package.
 	var minVersion uint16
@@ -213,11 +212,12 @@ func getTLSMinMaxVersionsClient(tlsConfig *s2av2pb.GetTlsConfigurationResp_Clien
 		maxVersion = tls.VersionTLS13
 	}
 	if minVersion > maxVersion {
-		return minVersion, maxVersion, errors.New("S2Av2 provided minVersion > maxVersion")
+		return minVersion, maxVersion, errors.New("S2Av2 provided minVersion > maxVersion.")
 	}
 	return minVersion, maxVersion, nil
 }
 
+// TODO(rmehta19): Test every possible case of input/outputs for this function.
 func getTLSMinMaxVersionsServer(tlsConfig *s2av2pb.GetTlsConfigurationResp_ServerTlsConfiguration) (uint16, uint16, error) {
 	// Map S2Av2 TLSVersion to consts defined in tls package.
 	var minVersion uint16
@@ -248,7 +248,7 @@ func getTLSMinMaxVersionsServer(tlsConfig *s2av2pb.GetTlsConfigurationResp_Serve
 		maxVersion = tls.VersionTLS13
 	}
 	if minVersion > maxVersion {
-		return minVersion, maxVersion, errors.New("S2Av2 provided minVersion > maxVersion")
+		return minVersion, maxVersion, errors.New("S2Av2 provided minVersion > maxVersion.")
 	}
 	return minVersion, maxVersion, nil
 }
--- a/internal/v2/tls_config_store/tls_config_store_test.go
+++ b/internal/v2/tls_config_store/tls_config_store_test.go
@@ -221,7 +221,7 @@ func TestTLSConfigStoreServer(t *testing.T) {
 	stop()
 }
 
-func TestgetTLSMinMaxVersionsClient(t *testing.T) {
+func TestGetTLSMinMaxVersionsClient(t *testing.T) {
 	var tlsconfig s2av2pb.GetTlsConfigurationResp_ClientTlsConfiguration
 	tlsconfig.MinTlsVersion = commonpb.TLSVersion_TLS_VERSION_1_0
 	tlsconfig.MaxTlsVersion = commonpb.TLSVersion_TLS_VERSION_1_1
@@ -235,7 +235,7 @@ func TestgetTLSMinMaxVersionsClient(t *testing.T) {
 	}
 }
 
-func TestgetTLSMinMaxVersionsServer(t *testing.T) {
+func TestGetTLSMinMaxVersionsServer(t *testing.T) {
 	var tlsconfig s2av2pb.GetTlsConfigurationResp_ServerTlsConfiguration
 	tlsconfig.MinTlsVersion = commonpb.TLSVersion_TLS_VERSION_1_0
 	tlsconfig.MaxTlsVersion = commonpb.TLSVersion_TLS_VERSION_1_1


internal/v2/tls_config_store/tls_config_store.go
```diff
@@ -21,54 +33,242 @@ var (
 	serverKey []byte
 )
 
+// TODO(rmehta19): Consider refactoring stream creation for client/server into
+// a helper function.
+
 // GetTlsConfigurationForClient returns a tls.Config instance for use by a client application.
 func GetTlsConfigurationForClient(serverHostname string) *tls.Config {
```
matthewstevenson88, line 40: I think the stream need to be passed to GetTlsConfigurationForClient rather than creating the stream inside the implementation.

internal/v2/tls_config_store/tls_config_store.go
```diff
@@ -3,14 +3,26 @@ package tlsconfigstore
...
+	s2av2pb "github.com/google/s2a-go/internal/proto/v2/s2a_go_proto"
+	commonpb "github.com/google/s2a-go/internal/proto/v2/common_go_proto"
+)
+
+const (
+	defaultTimeout = 10.0 * time.Second
 )
 
 var (
+	fakes2av2Address = flag.String("address", "0.0.0.0:8008", "Fake S2Av2 address")
```
matthewstevenson88, line 25: The variable (along with defaultTimeout) can be removed because we should only be passing the stream to the GetTlsConfigurationForClient API.

internal/v2/tls_config_store/tls_config_store.go
```diff
@@ -21,54 +33,242 @@ var (
...
+	if err != nil  {
+		log.Fatalf("Client: failed to setup bidirectional streaming RPC session: %v", err)
+	}
+	log.Printf("Client: set up bidirectional streaming RPC session.")
+
+	// Send request to S2Av2 for config.
+	err = cstream.Send(&s2av2pb.SessionReq {
+		AuthenticationMechanisms: []*s2av2pb.AuthenticationMechanism {
+			{
+				MechanismOneof: &s2av2pb.AuthenticationMechanism_Token{"token"},
```
matthewstevenson88, line 69: Please add a TODO to populate the authentication mechanism using the tokenmanager.

internal/v2/tls_config_store/tls_config_store.go
```diff
@@ -21,54 +33,242 @@ var (
...
+		maxVersion = tls.VersionTLS10
+	case commonpb.TLSVersion_TLS_VERSION_1_1:
+		maxVersion = tls.VersionTLS11
+	case commonpb.TLSVersion_TLS_VERSION_1_2:
+		maxVersion = tls.VersionTLS12
+	case commonpb.TLSVersion_TLS_VERSION_1_3:
+		maxVersion = tls.VersionTLS13
+	default:
+		maxVersion = tls.VersionTLS13
+	}
```
matthewstevenson88, line 141: Should we do a sanity check to make sure minVersion <= maxVersion?

internal/v2/tls_config_store/tls_config_store.go
```diff
@@ -21,54 +33,242 @@ var (
...
 
 	rootCertPool := x509.NewCertPool()
 	rootCertPool.AppendCertsFromPEM(serverCert)
 
-	// TODO(rmehta19): Call S2Av2 for config values.
+	// Map S2Av2 TLSVersion to consts defined in tls package.
+	var minVersion uint16
+	var maxVersion uint16
+	switch x := tlsConfig.MinTlsVersion; x {
+	case commonpb.TLSVersion_TLS_VERSION_UNSPECIFIED:
```
matthewstevenson88, line 114: Should we omit the "unspecified" case and just let it fall into the default bucket?

matthewstevenson88 (commented):
Hey Riya! I did a quick first pass. PTAL. :)

internal/v2/tls_config_store/tls_config_store.go
rmehta19: Good idea, thanks!

internal/v2/tls_config_store/tls_config_store.go
rmehta19: Will do, thanks!

internal/v2/tls_config_store/tls_config_store.go
rmehta19: Thanks for pointing this out! I moved the setup of the bidirectional stream to the client and server handshake, which call GetTlsConfigurationForClient/Server, and modified the tls config store unit store accordingly

rmehta19:
Thanks for the comments Matt!

internal/v2/s2av2.go
```diff
@@ -57,12 +68,35 @@ func (c *s2av2TransportCreds) ClientHandshake(ctx context.Context, serverAuthori
...
+		log.Fatalf("Client: failed to connect: %v", err)
+	}
+	defer conn.Close()
+	client := s2av2pb.NewS2AServiceClient(conn)
+	log.Printf("Client: connected to: %s", *fakes2av2Address)
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTimeout)
+	defer cancel()
+
+	// Setup bidrectional streaming session.
+	callOpts := []grpc.CallOption{}
```
matthewstevenson88, line 88: nit: I think we can omit the local variable and just directly pass []grpc.CallOption{} to the SetUpSession method.

internal/v2/tls_config_store/tls_config_store.go
```diff
@@ -22,53 +24,188 @@ var (
...
 	}
 
 	rootCertPool := x509.NewCertPool()
 	rootCertPool.AppendCertsFromPEM(serverCert)
 
-	// TODO(rmehta19): Call S2Av2 for config values.
+	// Map S2Av2 TLSVersion to consts defined in tls package.
+	var minVersion uint16
+	var maxVersion uint16
+	switch x := tlsConfig.MinTlsVersion; x {
```
matthewstevenson88, line 77: Can we move this switch statement into an unexported helper function?

internal/v2/tls_config_store/tls_config_store.go
```diff
@@ -22,53 +24,188 @@ var (
...
+	for _, v := range tlsConfig.CertificateChain {
+		// TODO(rmehta19): Call remote signer library for private key.
+		cert, err := tls.X509KeyPair([]byte(v), clientKey)
+		if err != nil {
+			log.Fatalf("Failed to generate X509KeyPair: %v", err)
+		}
+		certList = append(certList, cert)
 	}
 
 	rootCertPool := x509.NewCertPool()
```
matthewstevenson88, line 71: Per our discussion during yesterday's standup, I think we can remove the rootCertPool.

internal/v2/tls_config_store/tls_config_store.go
```diff
@@ -22,53 +24,188 @@ var (
...
+	case commonpb.TLSVersion_TLS_VERSION_1_2:
+		maxVersion = tls.VersionTLS12
+	case commonpb.TLSVersion_TLS_VERSION_1_3:
+		maxVersion = tls.VersionTLS13
+	default:
+		maxVersion = tls.VersionTLS13
+	}
+
+	if minVersion > maxVersion {
+		log.Printf("S2Av2 provided minVersion > maxVersion")
```
matthewstevenson88, line 195: Please return an error in this case.

internal/v2/tls_config_store/tls_config_store.go
```diff
@@ -22,53 +24,188 @@ var (
...
+		MaxVersion: maxVersion,
 	}
 }
 
 // GetTlsConfigurationForServer returns a tls.Config instance for use by a server application.
-func GetTlsConfigurationForServer() *tls.Config {
-	// TODO(rmehta19): Call S2Av2 for certificate.
-	// TODO(rmehta19): Call remote signer library for private key.
-	cert, err := tls.X509KeyPair(serverCert, serverKey)
+func GetTlsConfigurationForServer(cstream s2av2pb.S2AService_SetUpSessionClient) *tls.Config {
```
matthewstevenson88, line 119: When building the server-side tls.Config, we don't want to call S2A as soon as GetTlsConfigurationForServer is called. Instead, we want to create a helper function that will do so and use it to populate the tls.Config's GetConfigForClient field. We can make this change in a follow-up PR if you want, but please add a TODO as a reminder.

matthewstevenson88 (commented):
Thanks Riya! Here's another batch of comments. :)

internal/v2/tls_config_store/tls_config_store.go
```diff
@@ -22,53 +24,188 @@ var (
...
+
+	// Extract TLS configiguration from SessionResp.
+	tlsConfig := resp.GetGetTlsConfigurationResp().GetClientTlsConfiguration()
+
+	// Fake S2Av2 only puts one cert in CertitificateChain, for loop iterates
+	// once.
+	var certList []tls.Certificate
+	for _, v := range tlsConfig.CertificateChain {
+		// TODO(rmehta19): Call remote signer library for private key.
+		cert, err := tls.X509KeyPair([]byte(v), clientKey)
```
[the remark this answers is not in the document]
rmehta19: I see, I used a for loop here, because in the future we might have a chain of certs from leaf to root. For the fake implementation of the s2av2 service layer, the chain is only of length 1, which is why I hardcoded the key. I thought that in the future, in the for loop, we can make a call to the remote signer library for the private key corresponding to a cert(remote signer lib then calls s2av2 for this info) from the list.

Does s2av2 only maintain private key for the leaf cert in the certificate chain?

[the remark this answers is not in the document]
rmehta19: Also, another question, with these two if statements, if I'm understanding correctly, we will end up creating a cert with an empty (PrivateKey field)[https://pkg.go.dev/crypto/tls#Certificate]. Is it ok to populate the config with certs that don't have that field populated?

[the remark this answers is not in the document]
matthewstevenson88: Correct that the `GetTlsConfigurationResp` will, in general, consist of one or more certificates and that the S2A only maintains the private key for the leaf certificate.

RE the second point: this is my bad, I misunderstood how the API works. Apologies!

Let me expand on how I think this should work:

* We need to set the `Certificates` field of the `tls.Config`. We should populate it with an array containing exactly 1 instance of the `Certificate` struct.
* To create this instance of the `Certificate` struct, we need to populate the `Certificates` field and the `PrivateKey` field.
* To populate the `Certificates` field, we do the following: for each cert in the `GetTlsConfigurationResp`, we convert it from a PEM string to a DER byte array, and add this array to the `Certificates` field. I think this conversion can be done using the `pem.Decode` function, based on this documentation: https://pkg.go.dev/encoding/pem#Decode
* To populate the `PrivateKey` field, we should only use the private key corresponding to the leaf cert (in the short term), which we will later replace with the RemoteSigner.

Does that make sense slash answer the question?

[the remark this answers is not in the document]
rmehta19: Thanks for the detailed explanation Matt! Yes, this definitely clears up the problem I was having. I understood the S2Av2 gives us the certificate chain for the client / server, what I then (incorrectly) did is use the for loop to generate a new tls.Certificate for each cert in the chain, I then appended each certificate to list of []tls.Certificate, to pass to the config. After this explanation and today's standup, I understand that each cert in the chain S2Av2 gives us, will be parsed and appended to a singular tls.Certificate. 

The certificate decoding and parsing works using the pem.Decode and x509.Parsecertificate, the issue I faced after that was populating the PrivateKey field of the tls.Certificate. Since the Remote Signer fake is not yet implemented, I found this workaround:

Since the fake supplies the cert chain from leaf to root, for both the server and the client, I call:
`tlsCert, err := tls.X509KeyPair([]byte(tlsConfig.CertificateChain[0]), client/serverKey)`
`cert.PrivateKey = tlsCert.PrivateKey`
where `cert` is the tls.Certificate object we appended to in the for loop where we iterated through the S2Av2 provided cert chain. tlsCert is essentialy an intermediate variable, it serves no purpose other than the fact that I use its PrivateKey field for my cert object. I plan on removing this workaround after the Private Key operation fake is implented and called from the remote signer lib(which implements the crypto.Signer interface), which will be called from here. It's not the most elegant thing to generate a TLS cert only to use its PrivateKey field, but right now, I think it's the simplest. Does this sound ok? Do you think there's a better way to go about this?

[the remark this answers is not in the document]
rmehta19: Thanks for the detailed explanation Matt! Yes, this definitely clears up the problem I was having. I understood the S2Av2 gives us the certificate chain for the client / server, what I then (incorrectly) did is use the for loop to generate a new tls.Certificate for each cert in the chain, I then appended each certificate to list of []tls.Certificate, to pass to the config. After this explanation and today's standup, I understand that each cert in the chain S2Av2 gives us, will be parsed and appended to a singular tls.Certificate. 

The certificate decoding and parsing works using the pem.Decode and x509.Parsecertificate, the issue I faced after that was populating the PrivateKey field of the tls.Certificate. Since the Remote Signer fake is not yet implemented, I found this workaround:

Since the fake supplies the cert chain from leaf to root, for both the server and the client, I call:
`tlsCert, err := tls.X509KeyPair([]byte(tlsConfig.CertificateChain[0]), client/serverKey)`
`cert.PrivateKey = tlsCert.PrivateKey`
where `cert` is the tls.Certificate object we appended to in the for loop where we iterated through the S2Av2 provided cert chain. tlsCert is essentialy an intermediate variable, it serves no purpose other than the fact that I use its PrivateKey field for my cert object. I plan on removing this workaround after the Private Key operation fake is implented and called from the remote signer lib(which implements the crypto.Signer interface), which will be called from here. It's not the most elegant thing to generate a TLS cert only to use its PrivateKey field, but right now, I think it's the simplest. Does this sound ok? Do you think there's a better way to go about this?

internal/v2/tls_config_store/tls_config_store.go
rmehta19: Will do, I'll add a TODO and be sure to get to it in the next PR! This helper function will be something like 'sendServerTLSConfigRequest' and it will build the SessionReq, rather than doing that in the function itself? Do we want to do the same on the client side also then? Also, is the reasoning behind moving the cstream.Send(&SessionReq) to make the GetTlsConfiguration implementation simpler?

rmehta19: Will do, I'll add a TODO and be sure to get to it in the next PR! This helper function will be something like 'sendServerTLSConfigRequest' and it will build the SessionReq, rather than doing that in the function itself? Do we want to do the same on the client side also then? Also, is the reasoning behind moving the cstream.Send(&SessionReq) to make the GetTlsConfiguration implementation simpler?

internal/v2/s2av2.go
```diff
@@ -6,14 +6,25 @@ import (
 	"errors"
```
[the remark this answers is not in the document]
rmehta19: For this PR, I tested it by starting up the client, server, and fakes2av2 server binaries, I'll add screenshots(with and without the logging) in the description. I agree that it would be a good idea to add unit tests to test the Client and Server Handshakes in the success case; is it ok if I add this as a TODO and address in a follow up PR?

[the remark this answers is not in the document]
rmehta19: For this PR, I tested it by starting up the client, server, and fakes2av2 server binaries, I'll add screenshots(with and without the logging) in the description. I agree that it would be a good idea to add unit tests to test the Client and Server Handshakes in the success case; is it ok if I add this as a TODO and address in a follow up PR?

internal/v2/tls_config_store/tls_config_store.go
rmehta19: Is it ok to leave this in for now, and remove it once the VerifyPeerCertificateChain fake is implemented?

rmehta19: Is it ok to leave this in for now, and remove it once the VerifyPeerCertificateChain fake is implemented?

rmehta19:
Testing s2av2 with Logging

bazel run server (running go Echo server)
image

bazel run fakes2av2_server
image

bazel run client
image

rmehta19:
Testing s2av2 "without" Logging(retained the minimal amount of logging in the testonly code pre-This PR)

bazel run server (running go Echo server):
image

bazel run fakes2av2_server:
image

bazel run client:
image

internal/v2/tls_config_store/tls_config_store.go
matthewstevenson88: Sure that's fine. Let's just make sure to add a TODO so we don't forget to remove it. :)

internal/v2/tls_config_store/tls_config_store.go
```diff
@@ -22,53 +25,217 @@ var (
...
+		// Populate Certificates field
+		block, _ := pem.Decode([]byte(v))
+		if block == nil {
+			return nil, errors.New("certificate in CertificateChain obtained from S2Av2 is empty.")
+		}
+		x509Cert, err := x509.ParseCertificate(block.Bytes)
+		if err != nil {
+			return nil, err
+		}
+		cert.Certificate = append(cert.Certificate, x509Cert.Raw)
```
rmehta19, line 68: I just realized there is a "Leaf" field in the tls.Certificate (https://pkg.go.dev/crypto/tls#Certificate). I'm adding all the certs, including the leaf to the "Certificate". I think I should also add the leaf cert to the "Leaf" field.

internal/v2/tls_config_store/tls_config_store.go
rmehta19: Will do, thanks!

internal/v2/tls_config_store/tls_config_store.go
```diff
@@ -22,53 +25,217 @@ var (
...
 
 // GetTlsConfigurationForServer returns a tls.Config instance for use by a server application.
-func GetTlsConfigurationForServer() *tls.Config {
-	// TODO(rmehta19): Call S2Av2 for certificate.
-	// TODO(rmehta19): Call remote signer library for private key.
-	cert, err := tls.X509KeyPair(serverCert, serverKey)
+func GetTlsConfigurationForServer(cstream s2av2pb.S2AService_SetUpSessionClient) (*tls.Config, error) {
+	// TODO(rmehta19): move call to S2Av2 to a helper function.
+	// Send request to S2Av2 for config.
+	err := cstream.Send(&s2av2pb.SessionReq {
```
matthewstevenson88, line 106: micro-nit: here as well, please use the `if err := ...; err != nil { ... }` style

internal/v2/tls_config_store/tls_config_store_test.go
```diff
@@ -19,6 +32,23 @@ var (
 	serverKeypem []byte
 )
 
+func startFakeS2Av2Server(wg *sync.WaitGroup) (stop func(), err error, address string) {
```
matthewstevenson88, line 35: nit: the error should be the last thing returned in the list of outputs

internal/v2/tls_config_store/tls_config_store.go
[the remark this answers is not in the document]
rmehta19: Discussed offline, this is all good, added todo to move this private key stuff to remote signer.

internal/v2/tls_config_store/tls_config_store.go
rmehta19: Tested with echo/server, fakes2av2_server and echo/client, and it works after this change.

internal/v2/s2av2_test.go
```diff
@@ -1,12 +1,19 @@
...
 import (
+	"flag"
 	"testing"
 	"context"
 	"github.com/google/go-cmp/cmp"
 	"google.golang.org/protobuf/testing/protocmp"
 )
 
+var (
+	fakes2av2Address = flag.String("address", "0.0.0.0:8008", "Fake S2Av2 address")
```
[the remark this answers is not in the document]
rmehta19: The fakes2av2_server is set to run at :8008; in these unit tests, I don't setup a fake service because the tests never reach the point of actually dialing and creating a stream to fakes2av2, since the unit tests are only testing the failure case. Now that I think about it, I could just use :0 here, not hardcode the actually port the fakes2av2 server runs on, because if we ever do change the unit tests to test the success case of the handshake(actually dial and setup stream), then we will have toe startup and tear down a fake server in the test. 

Also, thanks for pointing it out, no need for it to be a flag. 

To summarize:
- I'll change the port to :0 
- I'lll make the port a global var, not a flag.

Thanks!

rmehta19:
Thanks for all the comments Matt! I made the changes and also started everything up again to verify it all works

internal/v2/tls_config_store/tls_config_store.go
```diff
@@ -22,53 +25,230 @@ var (
...
+		block, _ := pem.Decode([]byte(v))
+		if block == nil {
+			return nil, errors.New("certificate in CertificateChain obtained from S2Av2 is empty.")
+		}
+		x509Cert, err := x509.ParseCertificate(block.Bytes)
+		if err != nil {
+			return nil, err
+		}
+		cert.Certificate = append(cert.Certificate, x509Cert.Raw)
+		if i == lenChain {
```
matthewstevenson88, line 70: nit: shouldn't this be i==0?

internal/v2/tls_config_store/tls_config_store_test.go
```diff
@@ -114,4 +218,33 @@ func TestTLSConfigStoreServer(t *testing.T) {
 			}
 		})
 	}
+	stop()
+}
+
+func TestgetTLSMinMaxVersionsClient(t *testing.T) {
```
matthewstevenson88, line 224: nit: s/get/Get

Similarly below

matthewstevenson88, line 224: Can we add a TODO for the following: test every possible case of input/outputs for this function?

I don't want to block this PR on this minor issue, so it's totally fine to do this in a follow-up.

matthewstevenson88 (approved):
LGTM with a couple minor comments. Thanks for the great PR Riya! This PR was probably the most complex part of the whole project, so really great work getting it done.

internal/v2/tls_config_store/tls_config_store_test.go
rmehta19: Will do, thanks!

rmehta19:
Thanks Matt! I really appreciate all the helpful comments :)

PR #32, opened 2022-06-09
Improve fakes2av2 test suite

rmehta19:
Instead of starting up / stopping the service per unit test, set it up once before the for loop, and use it for all the tests.


matthewstevenson88 (approved):

rmehta19:
In the last push of the branch, committed after running git rebase main, to pick up the other changes that have already merged.

PR #34, opened 2022-06-09
Fake Implementation of ValidatePeerCertificateChain

rmehta19:
Implemented fake implementation of ValidatePeerCertificateChain for both a ClientPeer and a ServerPeer.

Modified existing cert_verifier library implementation to call (fake) s2av2

Additional notes:
- client_intermediate_cert.pem is a cert that is signed by client_root_cert.pem/client_root_key.pem. Similarly, server_intermedaite_cert.pem is signed by server_root_cert.pem/server_root_key.pem
- domain.ext is the configuration I used when creating the chain of certificates on client and server side. The only parameter that we care about (for now) is basicConstraints:CA=true which allows the ..._intermediate_... to sign the ..._leaf_... .
- Since ValidatePeerCertificateChainReq_ClientPeer and ValidatePeerCertificateChainReq_ServerPeer both have a CertificateChain field that requires the chain to be a list of DER encoded certificates, used this link:
Creating DER encoded cert from PEM cert: https://serverfault.com/questions/1006639/create-der-certificatekey-from-pem, to create DER encoded certificates from the existing PEM encoded certificates.
- After reading documentation: https://pkg.go.dev/crypto/tls#Config:~:text=VerifyPeerCertificate%20func(rawCerts%20%5B%5D%5B%5Dbyte%2C%20verifiedChains%20%5B%5D%5B%5D*x509.Certificate)%20error, I changed cert_verifier_test.go to always pass nil for verified chains, since that is what API will do when we disable normal verification.


internal/v2/fakes2av2/fakes2av2.go
```diff
@@ -112,3 +196,28 @@ func getTlsConfiguration(req *s2av2pb.GetTlsConfigurationReq) (*s2av2pb.SessionR
...
+		Status: &s2av2pb.Status {
+			Code: StatusCode,
+			Details: StatusDetails,
+		},
+		RespOneof: &s2av2pb.SessionResp_ValidatePeerCertificateChainResp {
+			&s2av2pb.ValidatePeerCertificateChainResp {
+				ValidationResult: ValidationResult,
+				ValidationDetails: ValidationDetails,
+				Context: &s2av2ctx.S2AContext{
+					/* TODO(rmehta19): Fill out context fields.
```
rmehta19, line 211: @matthewstevenson88 , we can talk offline about this, I was wondering which of the context fields we can expect S2A to fill out. From there, I can fill them out with defaults here, and handle them in cert_verifier.go

internal/v2/fakes2av2/fakes2av2.go
```diff
@@ -53,6 +64,79 @@ func (s *Server) SetUpSession(stream s2av2pb.S2AService_SetUpSessionServer) erro
...
 }
 
+// TODO(rmehta19): consider also returning an error with this function.
+func validatePeerCertificateChain(req *s2av2pb.ValidatePeerCertificateChainReq) (*s2av2pb.SessionResp, error) {
+	switch x := req.PeerOneof.(type) {
+	case *s2av2pb.ValidatePeerCertificateChainReq_ClientPeer_:
+		// TODO(rmehta19): Handle Verification Mode(UNSPECIFIED, SPIFFE, or CONNECT_TO_GOOGLE).
+		derCertChain := req.GetClientPeer().CertificateChain
+		if len(derCertChain) == 0 {
+			// TODO(rmehta19): Is it an error for peer to have empty cert chain(I think
```
rmehta19, line 74: Is an empty cert chain in the SessionReq an error that S2A will raise?

internal/v2/cert_verifier/cert_verifier_test.go
```diff
@@ -35,39 +80,74 @@ func TestVerifyClientCertChain(t *testing.T) {
...
+			rawCerts: [][]byte{clientDERCert,},
 			verifiedChains: [][]*x509.Certificate{{cert1.Leaf},{}},
 			err: nil,
 		},
 	}{
 		t.Run(tc.description, func(t *testing.T) {
-			VerifyPeerCertificateFunc := VerifyClientCertificateChain(tc.expectedCommonName, tc.pool)
-			if got, want := VerifyPeerCertificateFunc(tc.rawCerts, tc.verifiedChains), tc.err; got != want {
-				t.Errorf("Peer Certificate verification failed: %v", got)
+			log.Printf("----TEST CASE: Verify Client Cert Chain: %s----", tc.description)
```
tavishvaidya, line 103: IIRC, I don't think this is required since on failure t.Run will automatically tell which test case failed/succeeded.

internal/v2/cert_verifier/cert_verifier_test.go
```diff
@@ -35,39 +80,74 @@ func TestVerifyClientCertChain(t *testing.T) {
...
+			ctx, cancel := context.WithTimeout(context.Background(), defaultTimeout)
+			defer cancel()
+
+			// Setup bidrectional streaming session.
+			callOpts := []grpc.CallOption{}
+			cstream, err := c.SetUpSession(ctx, callOpts...)
+			if err != nil  {
+				log.Fatalf("Client: failed to setup bidirectional streaming RPC session: %v", err)
+			}
+			log.Printf("Client: set up bidirectional streaming RPC session.")
```
tavishvaidya, line 126: Should we get rid of this print statement?

internal/v2/cert_verifier/cert_verifier_test.go
```diff
@@ -35,39 +80,74 @@ func TestVerifyClientCertChain(t *testing.T) {
 	for _, tc := range []struct {
 		description              string
 		pool                     *x509.CertPool
-		expectedCommonName       string
 		rawCerts                 [][]byte
 		verifiedChains           [][]*x509.Certificate
-		err                      error
+		err			 error
```
tavishvaidya, line 85: Can we rename this to expectedErr for clarity?

internal/v2/cert_verifier/cert_verifier.go
```diff
@@ -3,64 +3,83 @@ package certverifier
...
-	"time"
+
+	s2av2pb "github.com/google/s2a-go/internal/proto/v2/s2a_go_proto"
 )
 
 // VerifyClientCertificateChain verifies a client's certificate chain against
 // the roots in the provided pool and checks that the common name in the
 // client's leaf certificate matches the expected common name.
-func VerifyClientCertificateChain(expectedCommonName string, pool *x509.CertPool) func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
+func VerifyClientCertificateChain(pool *x509.CertPool, cstream s2av2pb.S2AService_SetUpSessionClient) func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
```
matthewstevenson88, line 13: nit: we can omit the `pool` parameter

internal/v2/cert_verifier/cert_verifier.go
```diff
@@ -3,64 +3,83 @@ package certverifier
 
 import (
 	"crypto/x509"
-	"fmt"
-	"time"
+
+	s2av2pb "github.com/google/s2a-go/internal/proto/v2/s2a_go_proto"
 )
 
 // VerifyClientCertificateChain verifies a client's certificate chain against
```
matthewstevenson88, line 10: We need to update this paragraph as well.

internal/v2/cert_verifier/cert_verifier.go
```diff
@@ -3,64 +3,83 @@ package certverifier
...
+		// TODO(rmehta19): Parse response.
 		return nil
 	}
 }
 
 // VerifyServerCertificateChain verifies a servers' certificate chain against
 // the roots in the provided pool and checks that the common name in the
 // servers' leaf certificate matches the expected common name.
-func VerifyServerCertificateChain(expectedCommonName string, hostname string, pool *x509.CertPool) func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
+func VerifyServerCertificateChain(hostname string, pool *x509.CertPool, cstream s2av2pb.S2AService_SetUpSessionClient) func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
```
matthewstevenson88, line 51: nit: same here RE removing the `pool` parameter.

internal/v2/cert_verifier/cert_verifier.go
```diff
@@ -3,64 +3,83 @@ package certverifier
...
 
-		if cert.Subject.CommonName != expectedCommonName {
-			return fmt.Errorf("certificate had Common Name %q, expected %q", cert.Subject.CommonName, expectedCommonName)
-		}
+		// TODO(rmehta19): Parse response.
 		return nil
 	}
 }
 
 // VerifyServerCertificateChain verifies a servers' certificate chain against
```
matthewstevenson88, line 48: We need to update this paragraph as well.

internal/v2/cert_verifier/cert_verifier.go
```diff
@@ -3,64 +3,83 @@ package certverifier
...
+		if err := cstream.Send(&s2av2pb.SessionReq {
+			AuthenticationMechanisms: []*s2av2pb.AuthenticationMechanism {
+				{
+					// TODO(rmehta19): Populate Authentication Mechanism using tokenmanager.
+					MechanismOneof: &s2av2pb.AuthenticationMechanism_Token{"token"},
+				},
+			},
+			ReqOneof: &s2av2pb.SessionReq_ValidatePeerCertificateChainReq {
+				&s2av2pb.ValidatePeerCertificateChainReq {
+					Mode: s2av2pb.ValidatePeerCertificateChainReq_SPIFFE,
```
matthewstevenson88, line 25: Please make "connecting to Google" the default mode.

internal/v2/cert_verifier/cert_verifier.go
```diff
@@ -3,64 +3,83 @@ package certverifier
...
+					PeerOneof: &s2av2pb.ValidatePeerCertificateChainReq_ClientPeer_ {
+						&s2av2pb.ValidatePeerCertificateChainReq_ClientPeer {
+							CertificateChain: rawCerts,
```
matthewstevenson88, line 28: Can you confirm that the order of certs in rawCerts is what CertificateChain is expecting?

internal/v2/cert_verifier/cert_verifier.go
```diff
@@ -3,64 +3,83 @@ package certverifier
...
+		if err := cstream.Send(&s2av2pb.SessionReq {
+			AuthenticationMechanisms: []*s2av2pb.AuthenticationMechanism {
+				{
+					// TODO(rmehta19): Populate Authentication Mechanism using tokenmanager.
+					MechanismOneof: &s2av2pb.AuthenticationMechanism_Token{"token"},
+				},
+			},
+			ReqOneof: &s2av2pb.SessionReq_ValidatePeerCertificateChainReq {
+				&s2av2pb.ValidatePeerCertificateChainReq {
+					Mode: s2av2pb.ValidatePeerCertificateChainReq_SPIFFE,
```
matthewstevenson88, line 63: Please make "connecting to Google" the default mode.

internal/v2/cert_verifier/cert_verifier.go
```diff
@@ -3,64 +3,83 @@ package certverifier
...
+						},
+					},
+				},
+			},
+		}); err != nil {
+			return err
 		}
 
-		if _, err = cert.Verify(opts); err != nil {
+		// Get the response containing config from S2Av2.
```
matthewstevenson88, line 76: micro-nit: please remove "containing config"

internal/v2/cert_verifier/cert_verifier.go
```diff
@@ -3,64 +3,83 @@ package certverifier
...
```
matthewstevenson88, line 37: micro-nit: please remove "containing config"

internal/v2/cert_verifier/cert_verifier_test.go
```diff
@@ -35,39 +80,74 @@ func TestVerifyClientCertChain(t *testing.T) {
...
 			verifiedChains: nil,
-			err: nil,
+			err: errors.New("rpc error: code = Unknown desc = Client Peer Verification failed: client cert chain is empty."),
 		},
 		{
 			description: "static : non-empty certPool",
 			pool: certPool1,
-			expectedCommonName: "s2a_test_cert",
-			rawCerts: cert1.Certificate,
+			rawCerts: [][]byte{clientDERCert,},
```
matthewstevenson88, line 97: Do we have any tests where more than 1 cert is sent? If not, please add them. :)

internal/v2/cert_verifier/cert_verifier_test.go
```diff
@@ -35,39 +80,74 @@ func TestVerifyClientCertChain(t *testing.T) {
...
+			callOpts := []grpc.CallOption{}
+			cstream, err := c.SetUpSession(ctx, callOpts...)
+			if err != nil  {
+				log.Fatalf("Client: failed to setup bidirectional streaming RPC session: %v", err)
+			}
+			log.Printf("Client: set up bidirectional streaming RPC session.")
+
+			VerifyPeerCertificateFunc := VerifyClientCertificateChain(tc.pool, cstream)
+			got, want := VerifyPeerCertificateFunc(tc.rawCerts, tc.verifiedChains), tc.err
+			if ((got == nil) || (want == nil)) && (got != want) {
```
matthewstevenson88, line 130: I found this logic a little confusing - could we replace it with the following:

if want == nil {
   if got != nil {
      // error
   }
} else {
  if got == nil {
    // error
   }
   // compare error strings from got and want
}

wdyt?

matthewstevenson88: Similarly for the server side test.

internal/v2/fakes2av2/fakes2av2.go
```diff
@@ -53,6 +64,79 @@ func (s *Server) SetUpSession(stream s2av2pb.S2AService_SetUpSessionServer) erro
 	}
 }
 
+// TODO(rmehta19): consider also returning an error with this function.
```
matthewstevenson88, line 67: Is this TODO needed? Aren't we already returning an error?

internal/v2/fakes2av2/fakes2av2.go
```diff
@@ -53,6 +64,79 @@ func (s *Server) SetUpSession(stream s2av2pb.S2AService_SetUpSessionServer) erro
...
+	case *s2av2pb.ValidatePeerCertificateChainReq_ClientPeer_:
+		// TODO(rmehta19): Handle Verification Mode(UNSPECIFIED, SPIFFE, or CONNECT_TO_GOOGLE).
+		derCertChain := req.GetClientPeer().CertificateChain
+		if len(derCertChain) == 0 {
+			// TODO(rmehta19): Is it an error for peer to have empty cert chain(I think
+			// so, because we set ClientAuth as require request and verify)? If so,
+			// (server requires nonempty certificate chain) will error be raised
+			// for us by some existing API function way before this peer verification?
+			err := errors.New("Client Peer Verification failed: client cert chain is empty.")
+			return buildValidatePeerCertificateChainSessionResp(400, err.Error(), s2av2pb.ValidatePeerCertificateChainResp_FAILURE, err.Error()), err
```
matthewstevenson88, line 79: Why 400?

internal/v2/fakes2av2/fakes2av2.go
matthewstevenson88: Yup!

internal/v2/fakes2av2/fakes2av2.go
```diff
@@ -53,6 +64,79 @@ func (s *Server) SetUpSession(stream s2av2pb.S2AService_SetUpSessionServer) erro
...
+		rootCertPool := x509.NewCertPool()
+		if ok := rootCertPool.AppendCertsFromPEM(clientCert); ok != true {
+			err := errors.New("Client Peer Verification failed: S2Av2 could not obtain/parse roots.")
+			return buildValidatePeerCertificateChainSessionResp(500, err.Error(), s2av2pb.ValidatePeerCertificateChainResp_FAILURE, err.Error()), err
+		}
+
+		// Verify the leaf certificate.
+		opts := x509.VerifyOptions{
+			CurrentTime: time.Now(),
+			Roots: rootCertPool,
```
matthewstevenson88, line 92: We also need to set the intermediates based on the non-leaf certs in the chain.

internal/v2/fakes2av2/fakes2av2.go
matthewstevenson88: Is this an HTTP status code? In general, we should use gRPC status codes rather than HTTP status codes.

https://developers.google.com/maps-booking/reference/grpc-api/status_codes

internal/v2/fakes2av2/fakes2av2.go
```diff
@@ -53,6 +64,79 @@ func (s *Server) SetUpSession(stream s2av2pb.S2AService_SetUpSessionServer) erro
...
+			s := fmt.Sprintf("Client Peer Verification failed: %v", err)
+			return buildValidatePeerCertificateChainSessionResp(400, s, s2av2pb.ValidatePeerCertificateChainResp_FAILURE, s), err
+		}
+		if _, err := x509LeafCert.Verify(opts); err != nil {
+			s := fmt.Sprintf("Client Peer Verification failed: %v", err)
+			return buildValidatePeerCertificateChainSessionResp(400, s, s2av2pb.ValidatePeerCertificateChainResp_FAILURE, s), err
+		}
+		return buildValidatePeerCertificateChainSessionResp(200, "", s2av2pb.ValidatePeerCertificateChainResp_SUCCESS, "Client Peer Verification succeeded"), nil
+	case *s2av2pb.ValidatePeerCertificateChainReq_ServerPeer_:
+		// TODO(rmehta19): Handle Verification Mode(UNSPECIFIED, SPIFFE, or CONNECT_TO_GOOGLE).
```
matthewstevenson88, line 105: Let's just expect connecting-to-Google for the time being. :)

internal/v2/fakes2av2/fakes2av2.go
```diff
@@ -53,6 +64,79 @@ func (s *Server) SetUpSession(stream s2av2pb.S2AService_SetUpSessionServer) erro
...
+		derCertChain := req.GetServerPeer().CertificateChain
```
matthewstevenson88, line 106: Can we move the verification logic into a helper function? The code is a bit complex so I think doing so would be helpful. :)

internal/v2/fakes2av2/fakes2av2.go
matthewstevenson88: I think we should set all of them. Please let me know if you need test data that we can use to populate these fields.

matthewstevenson88 (commented):
Thanks for the PR Riya! Generally looks good, just a bunch of relatively minor comments.

internal/v2/cert_verifier/cert_verifier_test.go
rmehta19: Will do! I left this in from when I was testing, made it easier to go through the logs

internal/v2/cert_verifier/cert_verifier_test.go
rmehta19: Is it ok if I leave it in for now? It's helpful to see that when I am debugging a test. I can add a TODO to remove all the logging from the library tests after I'm done with adding tests to them

internal/v2/cert_verifier/cert_verifier.go
rmehta19: Do you mean add a function to verify rawCerts is indeed a list of DER encoded certs ordered from leaf to root?

CertificateChain expects the DER encoded certs to be ordered from leaf to root:
https://github.com/google/s2a-go/blob/23127301b4c006bbb890458e74ce23f8d64cf719/internal/proto/v2/s2a_go_proto/s2a.pb.go#L1579

internal/v2/cert_verifier/cert_verifier_test.go
rmehta19: Will do! I create a chain of length 2 : root_cert.der which is self signed, and leaf_cert.der which is signed by root_cert.der.

internal/v2/fakes2av2/fakes2av2.go
rmehta19: Yes we are, I must've missed deleting this line, thanks for catching this!

internal/v2/fakes2av2/fakes2av2.go
rmehta19: Yes, I was using HTTP codes, I'll switch over to the gRPC ones, thanks for pointing this out!

internal/v2/cert_verifier/cert_verifier_test.go
```diff
@@ -1,117 +1,208 @@
...
-			rawCerts: cert1.Certificate,
-			verifiedChains: [][]*x509.Certificate{{cert1.Leaf},{}},
-			err: nil,
+			description: "chain of length 1",
+			rawCerts: [][]byte{clientDERCert,},
+			expectedErr: nil,
+		},
+		{
+			description: "chain of length 2",
+			rawCerts: [][]byte{clientSubordinateDERCert, clientDERCert,},
```
matthewstevenson88, line 82: Shouldn't the certs be ordered from leaf to root?

internal/v2/fakes2av2/fakes2av2.go
```diff
@@ -42,7 +50,10 @@ func (s *Server) SetUpSession(stream s2av2pb.S2AService_SetUpSessionServer) erro
 		case *s2av2pb.SessionReq_OffloadResumptionKeyOperationReq:
 			// TODO(rmehta19): Implement fake.
 		case *s2av2pb.SessionReq_ValidatePeerCertificateChainReq:
-			// TODO(rmehta19): Implement fake.
+			if resp, err = validatePeerCertificateChain(req.GetValidatePeerCertificateChainReq()); err != nil {
```
matthewstevenson88, line 53: nit: s/resp/_

Since `resp` is unused at present.

internal/v2/fakes2av2/fakes2av2.go
```diff
@@ -53,6 +64,18 @@ func (s *Server) SetUpSession(stream s2av2pb.S2AService_SetUpSessionServer) erro
...
 
+func validatePeerCertificateChain(req *s2av2pb.ValidatePeerCertificateChainReq) (*s2av2pb.SessionResp, error) {
+	switch x := req.PeerOneof.(type) {
+	case *s2av2pb.ValidatePeerCertificateChainReq_ClientPeer_:
+		return verifyClientPeer(req)
+	case *s2av2pb.ValidatePeerCertificateChainReq_ServerPeer_:
+		return verifyServerPeer(req)
+	default:
+		err := errors.New(fmt.Sprintf("Peer Verification failed: invalid Peer type %T", x))
+		return buildValidatePeerCertificateChainSessionResp(400, err.Error(), s2av2pb.ValidatePeerCertificateChainResp_FAILURE, err.Error(),  &s2av2ctx.S2AContext{}), err
```
matthewstevenson88, line 75: nit: please remove `400` and use a gRPC status code instead.

internal/v2/fakes2av2/fakes2av2.go
```diff
@@ -112,3 +135,116 @@ func getTlsConfiguration(req *s2av2pb.GetTlsConfigurationReq) (*s2av2pb.SessionR
...
+		s := fmt.Sprintf("Server Peer Verification failed: %v", err)
+		return buildValidatePeerCertificateChainSessionResp(3, s, s2av2pb.ValidatePeerCertificateChainResp_FAILURE, s, &s2av2ctx.S2AContext{}), err
+	}
+	if _, err := x509LeafCert.Verify(opts); err != nil {
+		s := fmt.Sprintf("Server Peer Verification failed: %v", err)
+		return buildValidatePeerCertificateChainSessionResp(3, s, s2av2pb.ValidatePeerCertificateChainResp_FAILURE, s, &s2av2ctx.S2AContext{}), err
+	}
+
+	return buildValidatePeerCertificateChainSessionResp(0, "", s2av2pb.ValidatePeerCertificateChainResp_SUCCESS, "Server Peer Verification succeeded",&s2av2ctx.S2AContext{
+		LeafCertSpiffeId: "", // TODO(rmehta19)
```
matthewstevenson88, line 244: Is it ok if we send an empty context for now and come back and populate it later?

Rationale: We need to populate all of this data based on your test certs, but your test certs probably don't have a SPIFFE ID, URI or DNS SANs, so we may need to change them.

Similarly for the server-side context.

internal/v2/cert_verifier/cert_verifier.go
rmehta19: Discussed offline: go will always pass from leaf to root, and in unit tests I also pass from leaf to root. To clear up things I will
- rename the example cert files with more descriptive names
- create a README explaining how I generate the example certs
- add a test case where I pass certs from root to leaf and verify an error occurs
- add a test case testing the intermediates field of Verify

I'll probably not do all these things in this PR, but in follow up PR's.

internal/v2/cert_verifier/cert_verifier_test.go
```diff
@@ -1,117 +1,208 @@
...
+	clientDERCert []byte
+	//go:embed example_cert_key/server_cert.der
+	serverDERCert []byte
+	//go:embed example_cert_key/client_subordinate_cert.der
+	clientSubordinateDERCert []byte
+	//go:embed example_cert_key/server_subordinate_cert.der
+	serverSubordinateDERCert []byte
 )
 
+// TODO(rmehta19): remove logging.
```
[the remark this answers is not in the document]
rmehta19: I was thinking about removing the logging when the client stream to fake s2a is successfully created.

internal/v2/cert_verifier/cert_verifier_test.go
```diff
@@ -1,117 +1,208 @@
...
 		},
 		{
-			description: "static : non-empty certPool",
-			pool: certPool1,
-			expectedCommonName: "s2a_test_cert",
+			description: "chain of length 2",
 			hostname: "host",
-			rawCerts: cert1.Certificate,
-			verifiedChains: [][]*x509.Certificate{{cert1.Leaf},{}},
+			rawCerts: [][]byte{serverSubordinateDERCert, serverDERCert, },
```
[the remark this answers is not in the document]
rmehta19: Discussed offline:
- They are ordered from leaf to root here, I thought the word subordinate meant implied leaf, I'll fix the naming.

internal/v2/fakes2av2/fakes2av2.go
rmehta19: I think resp is used, since I send the resp after the switch statement, I declared it above. Regardless, I split up the call to the helper function and checking of the error into two separate lines, thanks!

internal/v2/fakes2av2/fakes2av2.go
rmehta19: Will do, thanks for the rationale!

internal/v2/fakes2av2/fakes2av2.go
rmehta19: Removing the context fields for now as noted in another comment made about adding context fields when we construct new certs with cnf values

rmehta19:
Thanks for all the comments Matt!

internal/v2/cert_verifier/cert_verifier.go
matthewstevenson88: SGTM. Thanks Riya!

internal/v2/cert_verifier/cert_verifier_test.go
```diff
@@ -1,117 +1,221 @@
...
-			rawCerts: cert1.Certificate,
-			verifiedChains: [][]*x509.Certificate{{cert1.Leaf},{}},
-			err: nil,
+			rawCerts: [][]byte{serverRootDERCert, },
+			expectedErr: nil,
+		},
+		{
+			description: "chain of length 2_correct",
+			hostname: "host",
+			rawCerts: [][]byte{serverIntermediateDERCert, serverRootDERCert, },
```
matthewstevenson88, line 170: Can we use `{serverLeafDERCert, serverIntermediateDERCert}` here?

rmehta19: After tracing through this again, rawCerts is passed to CertificateChain in the proto: https://github.com/google/s2a-go/blob/23127301b4c006bbb890458e74ce23f8d64cf719/internal/proto/v2/s2a_go_proto/s2a.pb.go#L1524

The CertificateChain field has to include certs all the way from leaf to root. I changed this to a chain of 3: leaf->intermediate->root in both the Client Peer test and Server Peer test.

To align with the above fact, I also added to the fake rpc implementation:
Since we are expecting the CertificateChain in the req to contain leaf to root inclusive, I now check that the root given to us in the certificate chain(last element in cert chain) matches one of the roots in S2A(hardcoded since this is a fake)

matthewstevenson88 (approved):

internal/v2/cert_verifier/cert_verifier_test.go
```diff
@@ -1,117 +1,221 @@
...
+			expectedErr: nil,
+		},
+		{
+			description: "chain of length 2_error",
+			rawCerts: [][]byte{serverLeafDERCert, serverRootDERCert,},
```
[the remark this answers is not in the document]
rmehta19: Discussed offline why this test case doesn't cause an error. Will remove it in this PR. To create another failure will create a chain of 4 certs in another PR.

internal/v2/cert_verifier/cert_verifier_test.go
rmehta19: Discussed offline:
- Correct documentation in proto file to make clear that CertificateChain does NOT include root
- Remove the extra check added to the fake RPC

PR #37, opened 2022-06-14
Move Private Key operation from tls_config_store to remote_signer

rmehta19:


tavishvaidya (approved):
Let's wait for @matthewstevenson88 to also review this change before merging this PR. Thanks!

internal/v2/fakes2av2/fakes2av2.go
```diff
@@ -98,15 +99,19 @@ func offloadPrivateKeyOperation(req *s2av2pb.OffloadPrivateKeyOperationReq, host
...
-		// should be used here. Currently hardcoded to use SHA256.
-
-		// Sign the SHA256 hash with the private key.
-		var opts crypto.Hash = crypto.SHA256
-		signedBytes, err := root.PrivateKey.(crypto.Signer).Sign(rand.Reader, req.GetInBytes(), opts)
-		if err != nil {
-			return nil, err
+		var signedBytes []byte
+		if req.GetSignatureAlgorithm() == s2av2pb.SignatureAlgorithm_S2A_SSL_SIGN_RSA_PKCS1_SHA256 {
+			var opts crypto.Hash = crypto.SHA256
```
matthewstevenson88, line 104: readability nit: I think this line should just be `opts := crypto.SHA256`

In fact, should we just omit the `opts` local variable altogether?

internal/v2/fakes2av2/fakes2av2.go
```diff
@@ -98,15 +99,19 @@ func offloadPrivateKeyOperation(req *s2av2pb.OffloadPrivateKeyOperationReq, host
...
+			signedBytes, err = root.PrivateKey.(crypto.Signer).Sign(rand.Reader, req.GetInBytes(), opts)
+			if err != nil {
+				return nil, err
+			}
+		} else if req.GetSignatureAlgorithm() == s2av2pb.SignatureAlgorithm_S2A_SSL_SIGN_RSA_PSS_RSAE_SHA256 {
```
matthewstevenson88, line 109: Should we add an "else" block to catch other signature algorithms and return an error in those cases?

internal/v2/fakes2av2/fakes2av2.go
```diff
@@ -98,15 +99,19 @@ func offloadPrivateKeyOperation(req *s2av2pb.OffloadPrivateKeyOperationReq, host
...
+			var opts = rsa.PSSOptions { -1, crypto.SHA256 }
```
matthewstevenson88, line 110: readability nit: same here, I think it should be `opts := &rsa.PSSOptions{SaltLength: -1, Hash: crypto.SHA256}`?

matthewstevenson88, line 110: For my own understanding, why is the salt length equal to -1?

internal/v2/fakes2av2/fakes2av2_test.go
```diff
@@ -253,7 +253,7 @@ func TestSetUpSession(t *testing.T) {
 				ReqOneof: &s2av2pb.SessionReq_OffloadPrivateKeyOperationReq {
 					&s2av2pb.OffloadPrivateKeyOperationReq {
 						Operation: s2av2pb.OffloadPrivateKeyOperationReq_SIGN,
-						SignatureAlgorithm: s2av2pb.SignatureAlgorithm_S2A_SSL_SIGN_RSA_PSS_RSAE_SHA256,
+						SignatureAlgorithm: s2av2pb.SignatureAlgorithm_S2A_SSL_SIGN_RSA_PKCS1_SHA256,
```
matthewstevenson88, line 256: Should we also add a test case that uses the s2av2pb.SignatureAlgorithm_S2A_SSL_SIGN_RSA_PSS_RSAE_SHA256 signature algorithm, so that we cover both codeblocks?

internal/v2/remote_signer/remote_signer.go
```diff
@@ -14,13 +14,14 @@ type remoteSigner struct {
 	leafCert *x509.Certificate
 	cstream s2av2pb.S2AService_SetUpSessionClient
 	localIdentity *commonpbv1.Identity
+	signatureAlgorithm s2av2pb.SignatureAlgorithm
```
matthewstevenson88, line 17: I don't think we can know the signatureAlgorithm in advance - we need to figure it out on every call to the `Sign` API. Since we're still not sure how to proceed with that (I think?), I would suggest the following for this PR: change the `getSignatureAlgorithm` method to accept the signerOpts and have it return a static value for now. WDYT?

internal/v2/remote_signer/remote_signer_test.go
```diff
@@ -96,19 +97,17 @@ func TestSign(t *testing.T) {
 	if err != nil {
 		log.Fatalf("failed to parse cert: %v", err)
 	}
+	testInBytes := []byte("Test data.")
```
matthewstevenson88, line 100: Please add a comment that explains why we're hashing the `testInBytes` (b/c we expect the real caller will have already done so).

internal/v2/remote_signer/remote_signer_test.go
```diff
@@ -96,19 +97,17 @@ func TestSign(t *testing.T) {
...
+
+	// Test RSA PKCS1v15 signature algorithm.
 	s := New(clientx509Cert, cstream, &commonpbv1.Identity {
 		IdentityOneof: &commonpbv1.Identity_Hostname {
 			Hostname: "client_hostname",
 		},
-	})
-	testInBytes := []byte("Test data.")
+	}, s2av2pb.SignatureAlgorithm_S2A_SSL_SIGN_RSA_PKCS1_SHA256)
+	var signerOpts crypto.Hash = crypto.SHA256
```
matthewstevenson88, line 109: readability nit: please replace with `signerOpts := crypto.SHA256`, or possibly even omit the local variable.

internal/v2/remote_signer/remote_signer_test.go
```diff
@@ -120,6 +119,22 @@ func TestSign(t *testing.T) {
...
+		t.Errorf("failed to verify RSA PKCS #1 v1.5 signature: %v", err)
+	}
+
+	// Test RSA PSS signature algorithm.
+	s = New(clientx509Cert, cstream, &commonpbv1.Identity {
+		IdentityOneof: &commonpbv1.Identity_Hostname {
+			Hostname: "client_hostname",
+		},
+	} , s2av2pb.SignatureAlgorithm_S2A_SSL_SIGN_RSA_PSS_RSAE_SHA256)
+	pssSignerOpts := rsa.PSSOptions {-1, crypto.SHA256 }
```
matthewstevenson88, line 132: readability nit: same here, I think it should be `pssSignerOpts := &rsa.PSSOptions{SaltLength: -1, Hash: crypto.SHA256}`?

internal/v2/tls_config_store/BUILD
```diff
@@ -17,7 +17,9 @@ go_library(
     "example_cert_key/server_key.pem",
   ],
   deps = [
+    "//internal/proto/common_go_proto:common_go_proto",
```
matthewstevenson88, line 20: Please group the deps alphabetically. :)

internal/v2/fakes2av2/fakes2av2.go
rmehta19: Sounds good! I'll pass crypto.SHA256 directly to the Sign function then.

internal/v2/fakes2av2/fakes2av2.go
rmehta19: Good point! I'll add this, and also add another unit test to in fakes2av2_test to catch this.

internal/v2/fakes2av2/fakes2av2.go
rmehta19: -1 is a constant: https://cs.opensource.google/go/go/+/refs/tags/go1.18.3:src/crypto/rsa/pss.go;l=231;bpv=0, perhaps I can use PSSSaltLengthEqualsHash, the named variable to make this more readable.

This is what the go crypto/tls calls the top - level Sign function(defined in remote_signer.go) with. For now, I've hardcoded it here, since this is the case for the go echo client and server binaries, during their client and server handshakes. However, perhaps there is a way we can retrieve this info(the opts variable) from the Client / Server Hello and pass it through to S2A through an additional proto field in OffloadPrivateKeyReq

internal/v2/fakes2av2/fakes2av2_test.go
rmehta19: I have a test case for this code block in the remote_signer_test.go file. 

The reason I didn't include it here is because of the structure of the tests in this file, we compare the proto responses. Since a PSS signature is not deterministic(salt is determined in the [SignPSS](https://cs.opensource.google/go/go/+/refs/tags/go1.18.3:src/crypto/rsa/pss.go;l=264;bpv=0) function), there isn't a way to compare the signed bytes directly. The way that I verify the PSS signature in remote_signer_test.go is by calling [VerifyPSS](https://cs.opensource.google/go/go/+/refs/tags/go1.18.3:src/crypto/rsa/pss.go;l=290;bpv=0). 

I could do the same thing I did in remote_signer_test.go here, by creating a separate function to verify this branch, to get full code coverage

internal/v2/remote_signer/remote_signer.go
rmehta19: Sounds good! I think there may be a way to get the signature algorithm info from the ClientHello information, if there is, we can make the change in the next PR.

internal/v2/fakes2av2/fakes2av2_test.go
matthewstevenson88: SG as long as its tested in remote_signer_test.go, that's fine.

rmehta19:
Thanks for the comments Matt and Tavish!

PR #40, opened 2022-06-15
[WIP] Test all possible cases in TestGetMinMaxVersionsClient/Server

rmehta19:


internal/v2/tls_config_store/tls_config_store_test.go
```diff
@@ -217,29 +218,41 @@ func TestTLSConfigStoreServer(t *testing.T) {
...
+			_, _, err := getTLSMinMaxVersionsServer(&tlsconfig)
+			if min <= max {
+				if err != nil {
+					t.Errorf("err = %v, expected err = nil", err)
+				}
+			} else {
+				if err == nil {
+					t.Errorf("err = nil, expected err = %v", errors.New("S2Av2 provided minVersion > maxVersion."))
+				}
+			}
```
tavishvaidya, line 255: Same here.

internal/v2/tls_config_store/tls_config_store_test.go
```diff
@@ -217,29 +218,41 @@ func TestTLSConfigStoreServer(t *testing.T) {
...
+		}
```
tavishvaidya, line 236: Let's test for the actual returned values rather than the relationship between the min and max values. Currently, the test may pass if the input values are 1.1 and 1.3 but the returned values are 1.2 and 1.3 for min and max versions. One way of doing this would be to create a mapping b/w the TLS lib versions and the proto versions and using that mapping in tests for expected vs got.

internal/v2/tls_config_store/tls_config_store.go
```diff
@@ -219,7 +218,6 @@ func getTLSMinMaxVersionsClient(tlsConfig *s2av2pb.GetTlsConfigurationResp_Clien
 	return minVersion, maxVersion, nil
 }
 
-// TODO(rmehta19): Test every possible case of input/outputs for this function.
 func getTLSMinMaxVersionsServer(tlsConfig *s2av2pb.GetTlsConfigurationResp_ServerTlsConfiguration) (uint16, uint16, error) {
 	// Map S2Av2 TLSVersion to consts defined in tls package.
 	var minVersion uint16
```
tavishvaidya, line 223: We should refactor the 4 switch statements to a helper function that takes the tlsConfig version and returns the tls.Versions. This switch statements can then be replaced with calls to this helper function. We can do this in a separate PR. Thanks!

rmehta19: Will do! I added a TODO.

tavishvaidya (approved):

internal/v2/tls_config_store/tls_config_store_test.go
```diff
@@ -216,30 +217,65 @@ func TestTLSConfigStoreServer(t *testing.T) {
...
-	tlsconfig.MinTlsVersion = commonpb.TLSVersion_TLS_VERSION_1_0
-	tlsconfig.MaxTlsVersion = commonpb.TLSVersion_TLS_VERSION_1_1
-	if _, _, err := getTLSMinMaxVersionsClient(&tlsconfig); err != nil {
-		t.Errorf("error in getTlsMinMaxVersionsClient logic: %v", err)
-	}
-	tlsconfig.MinTlsVersion = commonpb.TLSVersion_TLS_VERSION_1_1
-	tlsconfig.MaxTlsVersion = commonpb.TLSVersion_TLS_VERSION_1_0
-	if _, _, err := getTLSMinMaxVersionsClient(&tlsconfig); err == nil {
-		t.Errorf("getTlsMinMaxVersionsClient returned nil, should have returned error")
+	m := makeMap()
```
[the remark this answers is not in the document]
matthewstevenson88: Sorry for being unclear! What I was thinking was to have a test cases where we use:

1. `tlsConfig := &s2av2pb.GetTlsConfigurationResp_ServerTlsConfiguration {
				MinTlsVersion: commonpb.TLSVersion_TLS_VERSION_1_0 - 1,
				MaxTlsVersion: commonpb.TLSVersion_TLS_VERSION_1_3
			}`
2. `tlsConfig := &s2av2pb.GetTlsConfigurationResp_ServerTlsConfiguration {
				MinTlsVersion: commonpb.TLSVersion_TLS_VERSION_1_0,
				MaxTlsVersion: commonpb.TLSVersion_TLS_VERSION_1_3 + 1
			}`

Both of these examples should lead to errors. The rationale is: 1 is the defence-in-depth against the S2A returning a min version that is too low/invalid, and 2 is defence-in-depth against the S2A returning a max version that is too high/invalid.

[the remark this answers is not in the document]
rmehta19: I see, thanks for elaborating! To catch this, I'll change the current logic of the function, to throw an error in the default case rather than just setting to tls.VersionTLS13. I'll also add a new TestXXX to test this scenario for each getTLSMinMaxVersionsClient/Server

matthewstevenson88 (approved):

PR #42, opened 2022-06-17
Remote Signer makes RPC to Fake S2Av2

rmehta19:
Modify Remote Signer Library to make calls to fake s2a for signing.


internal/v2/remote_signer/remote_signer.go
```diff
@@ -2,13 +2,77 @@
...
 import (
+	"io"
 	"crypto"
-	"crypto/tls"
+	"crypto/x509"
+	s2av2pb "github.com/google/s2a-go/internal/proto/v2/s2a_go_proto"
+	commonpbv1 "github.com/google/s2a-go/internal/proto/common_go_proto"
 )
 
+type RemoteSigner struct {
```
matthewstevenson88, line 12: nit: I think this can be unexported.

Please add a comment anyway though. :)

internal/v2/remote_signer/remote_signer.go
```diff
@@ -2,13 +2,77 @@
...
+	leafCert *x509.Certificate
+	cstream s2av2pb.S2AService_SetUpSessionClient
+	hostname string
```
matthewstevenson88, line 15: What hostname is this? Should we maybe be storing the local identity instead?

If so, I think it should be:
```
cpb "github.com/google/s2a-go/internal/proto/common_go_proto"

localIdentity *cpb.Identity
```

internal/v2/remote_signer/remote_signer.go
```diff
@@ -2,13 +2,77 @@
...
+}
+
+func GetCert(s RemoteSigner) *x509.Certificate {
```
matthewstevenson88, line 18: Please add a comment for this function.

internal/v2/remote_signer/remote_signer.go
```diff
@@ -2,13 +2,77 @@
...
+	return s.hostname
+}
+
 // New returns an instance of RemoteSigner, an implementation of the
 // crypto.Signer interface.
-func New(cert *tls.Certificate) crypto.Signer {
-	// TODO(rmehta19): Implement crypto.Signer interface that calls S2Av2.
-	return cert.PrivateKey.(crypto.Signer)
+func New(leafCert *x509.Certificate, cstream s2av2pb.S2AService_SetUpSessionClient, hostname string) crypto.Signer {
+	return RemoteSigner{leafCert, cstream, hostname}
```
matthewstevenson88, line 33: nit: please return a reference to the RemoteSigner instead.

internal/v2/remote_signer/remote_signer.go
matthewstevenson88, line 18: Should this function be unexported? If so, can we move it to the bottom of the file.

internal/v2/remote_signer/remote_signer.go
```diff
@@ -2,13 +2,77 @@
...
+	leafCert *x509.Certificate
+	cstream s2av2pb.S2AService_SetUpSessionClient
+	hostname string
+}
+
+func GetCert(s RemoteSigner) *x509.Certificate {
+	return s.leafCert
+}
+
+func GetStream(s RemoteSigner) s2av2pb.S2AService_SetUpSessionClient {
```
matthewstevenson88, line 22: Should this function be unexported? If so, can we move it to the bottom of the file.

internal/v2/remote_signer/remote_signer.go
```diff
@@ -2,13 +2,77 @@
...
+	return s.cstream
+}
+
+func GetHostname(s RemoteSigner) string {
```
matthewstevenson88, line 26: Should this function be unexported? If so, can we move it to the bottom of the file.

internal/v2/remote_signer/remote_signer.go
matthewstevenson88, line 22: Please add a comment for this function.

internal/v2/remote_signer/remote_signer.go
matthewstevenson88, line 26: Please add a comment for this function.

matthewstevenson88, line 26: Should this maybe be GetLocalIdentity instead? Per comment above.

internal/v2/remote_signer/remote_signer.go
matthewstevenson88, line 18: I think each of these functions should be "member functions" of the struct, so the signature should be e.g.
```
func (s *RemoteSigner) GetCert() *x509.Certificate
``

File: internal/v2/remote_signer/remote_signer.go [hunk 7]
```diff
@@ -2,13 +2,77 @@
...
 // New returns an instance of RemoteSigner, an implementation of the
 // crypto.Signer interface.
-func New(cert *tls.Certificate) crypto.Signer {
-	// TODO(rmehta19): Implement crypto.Signer interface that calls S2Av2.
-	return cert.PrivateKey.(crypto.Signer)
+func New(leafCert *x509.Certificate, cstream s2av2pb.S2AService_SetUpSessionClient, hostname string) crypto.Signer {
+	return RemoteSigner{leafCert, cstream, hostname}
+}
+
+func (s RemoteSigner) Public() crypto.PublicKey {
```
matthewstevenson88, line 36: nit: s/RemoteSigner/*RemoteSigner

Similarly for Sign.

internal/v2/remote_signer/remote_signer_test.go
```diff
@@ -1,28 +1,173 @@
...
+
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+	"github.com/google/s2a-go/internal/v2/fakes2av2"
+	s2av2pb "github.com/google/s2a-go/internal/proto/v2/s2a_go_proto"
 )
 
+const (
+	defaultTimeout = 10.0 * time.Second
+)
```
matthewstevenson88, line 26: nit: please add a space between const () and the func

internal/v2/remote_signer/remote_signer_test.go
```diff
@@ -1,28 +1,173 @@
...
+	log.Printf("Client: set up bidirectional streaming RPC session.")
+
+	// Setup data for testing New.
+	clientx509Cert, err := x509.ParseCertificate(clientCertDER)
+	if err != nil {
+		t.Errorf("failed to parse cert: %v", err)
+	}
+
+	// Test New.
+	got := New(clientx509Cert, cstream, "client_hostname")
```
matthewstevenson88, line 162: We're not really using that this is a real stream - should we just pass in a fake stream instead?

internal/v2/remote_signer/remote_signer.go
rmehta19: This is the identity we're tying to SessionReq in the Sign API. That's a good idea, I'll change it to this, thanks!

internal/v2/remote_signer/remote_signer.go
```diff
@@ -2,13 +2,78 @@
...
+	localIdentity *commonpbv1.Identity
+}
+
+
 // New returns an instance of RemoteSigner, an implementation of the
 // crypto.Signer interface.
-func New(cert *tls.Certificate) crypto.Signer {
-	// TODO(rmehta19): Implement crypto.Signer interface that calls S2Av2.
-	return cert.PrivateKey.(crypto.Signer)
+func New(leafCert *x509.Certificate, cstream s2av2pb.S2AService_SetUpSessionClient, localIdentity *commonpbv1.Identity) *remoteSigner {
```
matthewstevenson88, line 22: Can we leave the return type as `crypto.Signer`?

matthewstevenson88 (approved):

PR #43, opened 2022-06-21
Populate ClientCAs field in Server TLS config

rmehta19:
This PR achieves a gRPC Go echo client and server that uses an actual implementation of TransportCredentials interface to establish an mTLS connection, by populating ClientCAs locally, in the server config, in the tls config store library. 

To elaborate on the TODO in the code, the crypto/tls [ClientAuth](https://cs.opensource.google/go/go/+/refs/tags/go1.18.3:src/crypto/tls/common.go;l=634) field in the crypto/tls [Config](https://cs.opensource.google/go/go/+/refs/tags/go1.18.3:src/crypto/tls/common.go;l=518) has several values:
(1) [NoClientCert](https://cs.opensource.google/go/go/+/refs/tags/go1.18.3:src/crypto/tls/common.go;l=297;drc=635b1244aa7671bcd665613680f527452cac7555;bpv=0;bpt=1)
(2) [RequestClientCert](https://cs.opensource.google/go/go/+/refs/tags/go1.18.3:src/crypto/tls/common.go;l=301;drc=635b1244aa7671bcd665613680f527452cac7555;bpv=0;bpt=1)
(3) [RequireAnyClientCert](https://cs.opensource.google/go/go/+/refs/tags/go1.18.3:src/crypto/tls/common.go;l=305;drc=635b1244aa7671bcd665613680f527452cac7555;bpv=0;bpt=1)
(4) [VerifyClientCertIfGiven](https://cs.opensource.google/go/go/+/refs/tags/go1.18.3:src/crypto/tls/common.go;l=310;drc=635b1244aa7671bcd665613680f527452cac7555;bpv=0;bpt=1)
(5) [RequireAndVerifyClientCert](https://cs.opensource.google/go/go/+/refs/tags/go1.18.3:src/crypto/tls/common.go;l=314;drc=635b1244aa7671bcd665613680f527452cac7555;bpv=0;bpt=1)

As per the documentation for [VerifyPeerCertificate](https://cs.opensource.google/go/go/+/refs/tags/go1.18.3:src/crypto/tls/common.go;l=600;drc=635b1244aa7671bcd665613680f527452cac7555;bpv=0;bpt=1), if in the server config, ClientAuth is set to RequestClientCert or RequireAnyClientCert, crypto/tls skips normal verification. If ClientAuth is set to VerifyClientCertIfGiven or RequireAndVerifyClientCert, crypto/tls performs normal verification before calling the function given in VerifyPeerCertificate. The crypto/tls normal verification requires that ClientCAs be populated, if not, then normal verification failed, the VerifyPeerCertificate custom implementation never gets called, and the server tls handshake fails. 

With our use of S2Av2 for the populating the server TLS config there are some things we can do;
- Never populate ClientAuth with a value that triggers the crypto/tls normal verification process(always set it to RequestClientCert or RequireAnyClientCert). VerifyPeerCertificate will still be called in this case, but the verifiedchains parameters will be nil.
- Add a field called ClientCAs to the [ServerTLSConfiguration](https://github.com/google/s2a-go/blob/main/internal/proto/v2/s2a.proto#L106), and use it in the client libraries, whenever S2Av2 sets [request_client_certificate](https://github.com/google/s2a-go/blob/main/internal/proto/v2/s2a.proto#L146) to REQUEST_AND_VERIFY.

### DIFF
--- a/internal/v2/tls_config_store/tls_config_store.go
+++ b/internal/v2/tls_config_store/tls_config_store.go
@@ -157,12 +157,25 @@ func GetTlsConfigurationForServer(cstream s2av2pb.S2AService_SetUpSessionClient)
 		return nil, err
 	}
 
+	// TODO(rmehta19): Choose (1) or (2)
+	// (1) Remove ClientCAs field if S2Av2 will always give
+	// tls.ClientAuthType = tls.NoClientCert, tls.RequestClientCert or
+	// tls.RequireAnyClientCert. For these 3 ClientAuthType's, go crypto/tls
+	// skips normal verification, and directly calls VerifyPeerCertificate func.
+	// (2) Get ClientCAs field from S2Av2, if S2Av2 gives tls.ClientAuthType =
+	// tls.VerifyClientCertIfGiven or tls.RequireAndVerifyClientCert.
+	// For these 2 ClientAuthType's, go crypto/tls performs normal verification,
+	// which requires ClientCAs, before custom VerifyPeerCertificate func.
+	clientCApool := x509.NewCertPool()
+	clientCApool.AppendCertsFromPEM(clientCert)
+
 	// Create mTLS credentials for server.
 	return &tls.Config {
 		// TODO(rmehta19): Make use of tlsConfig.HandshakeCiphersuites /
 		// RecordCiphersuites / TlsResumptionEnabled / MaxOverheadOfTicketAead.
 		Certificates: []tls.Certificate{cert},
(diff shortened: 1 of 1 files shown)


rmehta19:
Go echo client / server using actual implementation of transport credentials to establish an mTLS connection:

bazel run fakes2av2_server
image

bazel run server
image

bazel run client 
image

tavishvaidya (approved):
Thank you for clearly documenting the TODO and providing all the necessary context!

PR #44, opened 2022-06-21
Remove InsecureSkipVerify from server TLS config

rmehta19:
Remove InsecureSkipVerify from server TLS config.

As noted in the [documentation for InsecureSkipVerify](https://cs.opensource.google/go/go/+/refs/tags/go1.18.3:src/crypto/tls/common.go;l=647), this field is only used if set in the client TLS config. It controls whether a client verifies the server's certificate chain and host name.

Brought up Go echo client server to verify this change does not impact anything:

bazel run fakes2av2_server
image

bazel run server
image

bazel run client
image


tavishvaidya (approved):

tavishvaidya (approved):
Thanks for the fix!

PR #45, opened 2022-06-22
Handle SessionResp Status field in TLS Configuration Store

rmehta19:


internal/v2/tls_config_store/tls_config_store.go
```diff
@@ -49,7 +49,9 @@ func GetTlsConfigurationForClient(serverHostname string, cstream s2av2pb.S2AServ
 		return nil, err
 	}
 
-	// TODO(rmehta19): Handle resp.GetStatus().
+	if resp.GetStatus().Code != 0 {
```
tavishvaidya, line 52: Let's use codes.OK instead of the numeric value, makes it easier to read :)

internal/v2/tls_config_store/tls_config_store.go
```diff
@@ -49,7 +49,9 @@ func GetTlsConfigurationForClient(serverHostname string, cstream s2av2pb.S2AServ
...
+		return nil, errors.New(resp.GetStatus().Details)
+	}
```
tavishvaidya, line 54: While we are at it, let's also add a unit test to cover this if branch. This would correspond to any internal error returned by the S2A service to the S2A client.

rmehta19: Will do! I added a test for this branch in the fakes2av2_test suite

tavishvaidya (approved):

internal/v2/fakes2av2/fakes2av2.go
```diff
@@ -193,8 +193,13 @@ func getTlsConfiguration(req *s2av2pb.GetTlsConfigurationReq) (*s2av2pb.SessionR
 			},
 		}, nil
 	} else {
-		err := fmt.Errorf("unknown ConnectionSide, req.GetConnectionSide() returned %v", req.GetConnectionSide())
-		return nil, err
+		s := fmt.Sprintf("unknown ConnectionSide, req.GetConnectionSide() returned %v", req.GetConnectionSide())
+		return &s2av2pb.SessionResp {
+			Status: &s2av2pb.Status {
+				Code: 3,
```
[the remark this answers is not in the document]
rmehta19: Will do, I'll add a todo for now, and change all the Code fields in this file to use go grpc codes package. As discussed offlined, even though this means I need to cast: uint32(codes.InvalidArgument), it is still more readable than just listing 3.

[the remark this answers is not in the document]
rmehta19: Will do! To make merge conflict resolution a bit easier, I'm planning on changing all the status codes to use the named constants in PR#48(planning to merge that one last in this batch of ~9 open PRs)

internal/v2/tls_config_store/tls_config_store.go
```diff
@@ -49,7 +51,9 @@ func GetTlsConfigurationForClient(serverHostname string, cstream s2av2pb.S2AServ
 		return nil, err
 	}
 
-	// TODO(rmehta19): Handle resp.GetStatus().
+	if resp.GetStatus().Code != uint32(codes.OK) {
```
matthewstevenson88, line 54: Sorry I just noticed this: can we hit this if statement in a unit test?

If this PR is blocking other PRs in the queue, I'm fine adding a TODO for now and doing this in a follow-up.

Similarly below.

rmehta19: Will do! 

I think this will involve catching an error somewhere else on the fakes2av2 side. Currently fakes2av2 only populates the status with a non OK status when the connection side is invalid, the connection side is something that get's populated in the GetTlsConfigurationForClient/Server. I have a unit test in fakes2av2_test that tests this  already, but it directly makes the RPC with an invalid connection side rather than calling GetTlsConfigurationForClient/Server which makes the RPC.

For now, I can add to fakes2av2: read the SNI in GetTlsConfigurationReq, if SNI == "throw_internal_error", then throw an internal error. This will get us code coverage for this line, does that sound ok? I will add a TODO to get to this in another PR, since it involves more files changed than just these.

matthewstevenson88: SGTM! Thanks!

matthewstevenson88 (approved):

PR #46, opened 2022-06-22
Call the token manager for authentication mechanism

rmehta19:
Modify s2av2 to call tokenmanager library to establish a default token that can be attached to all RPCs made to s2av2. 

Note that in actual s2av2 implementation the only time the token is read is when the first SessionReq is made, all following SessionReq RPC authentication mechanisms are not read.

### DIFF
--- a/internal/v2/BUILD
+++ b/internal/v2/BUILD
@@ -13,7 +13,9 @@ go_library(
   importpath = "github.com/google/s2a-go/internal/v2",
   deps = [
     "//internal/handshaker/service",
+    "//internal/proto/common_go_proto:common_go_proto",
     "//internal/proto/v2/s2a_go_proto:s2a_go_proto",
+    "//internal/tokenmanager",
     "//internal/v2/tls_config_store",
     "@org_golang_google_grpc//credentials:go_default_library",
     "@org_golang_google_grpc//:go_default_library",
--- a/internal/v2/cert_verifier/cert_verifier.go
+++ b/internal/v2/cert_verifier/cert_verifier.go
@@ -15,12 +15,6 @@ func VerifyClientCertificateChain(cstream s2av2pb.S2AService_SetUpSessionClient)
 	return func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
 		// Offload verification to S2Av2.
 		if err := cstream.Send(&s2av2pb.SessionReq {
-			AuthenticationMechanisms: []*s2av2pb.AuthenticationMechanism {
-				{
-					// TODO(rmehta19): Populate Authentication Mechanism using tokenmanager.
-					MechanismOneof: &s2av2pb.AuthenticationMechanism_Token{"token"},
-				},
-			},
 			ReqOneof: &s2av2pb.SessionReq_ValidatePeerCertificateChainReq {
 				&s2av2pb.ValidatePeerCertificateChainReq {
 					Mode: s2av2pb.ValidatePeerCertificateChainReq_CONNECT_TO_GOOGLE,
@@ -56,12 +50,6 @@ func VerifyServerCertificateChain(hostname string, cstream s2av2pb.S2AService_Se
 	return func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
 		// Offload verification to S2Av2.
 		if err := cstream.Send(&s2av2pb.SessionReq {
-			AuthenticationMechanisms: []*s2av2pb.AuthenticationMechanism {
-				{
-					// TODO(rmehta19): Populate Authentication Mechanism using tokenmanager.
-					MechanismOneof: &s2av2pb.AuthenticationMechanism_Token{"token"},
-				},
-			},
 			ReqOneof: &s2av2pb.SessionReq_ValidatePeerCertificateChainReq {
 				&s2av2pb.ValidatePeerCertificateChainReq {
 					Mode: s2av2pb.ValidatePeerCertificateChainReq_CONNECT_TO_GOOGLE,
--- a/internal/v2/cert_verifier/cert_verifier_test.go
+++ b/internal/v2/cert_verifier/cert_verifier_test.go
@@ -62,7 +62,7 @@ func TestVerifyClientCertChain(t *testing.T) {
 	stop, address, err := startFakeS2Av2Server(&wg)
 	wg.Wait()
 	if err != nil {
-		log.Fatalf("error starting fake S2Av2 Server: %v", err)
+		t.Fatalf("error starting fake S2Av2 Server: %v", err)
 	}
 
 	for _, tc := range []struct {
@@ -100,7 +100,7 @@ func TestVerifyClientCertChain(t *testing.T) {
 			}
 			conn, err := grpc.Dial(address, opts...)
 			if err != nil {
-				log.Fatalf("Client: failed to connect: %v", err)
+				t.Fatalf("Client: failed to connect: %v", err)
 			}
 			defer conn.Close()
 			c := s2av2pb.NewS2AServiceClient(conn)
@@ -112,7 +112,7 @@ func TestVerifyClientCertChain(t *testing.T) {
 			callOpts := []grpc.CallOption{}
 			cstream, err := c.SetUpSession(ctx, callOpts...)
 			if err != nil  {
-				log.Fatalf("Client: failed to setup bidirectional streaming RPC session: %v", err)
+				t.Fatalf("Client: failed to setup bidirectional streaming RPC session: %v", err)
 			}
 			log.Printf("Client: set up bidirectional streaming RPC session.")
 
@@ -143,7 +143,7 @@ func TestVerifyServerCertChain(t *testing.T) {
 	stop, address, err := startFakeS2Av2Server(&wg)
 	wg.Wait()
 	if err != nil {
-		log.Fatalf("error starting fake S2Av2 Server: %v", err)
+		t.Fatalf("error starting fake S2Av2 Server: %v", err)
 	}
 
 	for _, tc := range []struct {
@@ -180,7 +180,7 @@ func TestVerifyServerCertChain(t *testing.T) {
 			}
 			conn, err := grpc.Dial(address, opts...)
 			if err != nil {
-				log.Fatalf("Client: failed to connect: %v", err)
+				t.Fatalf("Client: failed to connect: %v", err)
 			}
 			defer conn.Close()
 			c := s2av2pb.NewS2AServiceClient(conn)
@@ -192,7 +192,7 @@ func TestVerifyServerCertChain(t *testing.T) {
 			callOpts := []grpc.CallOption{}
 			cstream, err := c.SetUpSession(ctx, callOpts...)
 			if err != nil  {
-				log.Fatalf("Client: failed to setup bidirectional streaming RPC session: %v", err)
+				t.Fatalf("Client: failed to setup bidirectional streaming RPC session: %v", err)
 			}
 			log.Printf("Client: set up bidirectional streaming RPC session.")
 
--- a/internal/v2/s2av2.go
+++ b/internal/v2/s2av2.go
@@ -12,8 +12,10 @@ import (
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
 	"github.com/google/s2a-go/internal/v2/tls_config_store"
+	"github.com/google/s2a-go/internal/tokenmanager"
 	"github.com/google/s2a-go/internal/handshaker/service"
 	s2av2pb "github.com/google/s2a-go/internal/proto/v2/s2a_go_proto"
+	commonpbv1 "github.com/google/s2a-go/internal/proto/common_go_proto"
 )
 
 const (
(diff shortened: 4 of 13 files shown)


internal/v2/s2av2.go
```diff
@@ -69,14 +71,24 @@ func (c *s2av2TransportCreds) ClientHandshake(ctx context.Context, serverAuthori
 	if err != nil {
 		return nil, nil, err
 	}
+	// Create an AccessTokenManager instance to use to authenticate to S2Av2.
+	os.Setenv("S2A_ACCESS_TOKEN", "token")
```
matthewstevenson88, line 75: We don't want to set this environment variable in the actual source code - it may or may not be set by the Cloud infrastructure depending on where the application is running.

internal/v2/s2av2.go
```diff
@@ -69,14 +71,24 @@ func (c *s2av2TransportCreds) ClientHandshake(ctx context.Context, serverAuthori
...
+	accessTokenManager, err := tokenmanager.NewSingleTokenAccessTokenManager()
```
matthewstevenson88, line 76: Can we store the accessTokenManager as a field in the s2av2TransportCreds struct, so that we're not creating a new one for each TLS connection?

internal/v2/s2av2.go
```diff
@@ -69,14 +71,24 @@ func (c *s2av2TransportCreds) ClientHandshake(ctx context.Context, serverAuthori
...
+		return nil, nil, err
+	}
+	token, err := accessTokenManager.DefaultToken()
+	if err != nil {
+		return nil, nil, err
+	}
 	var config *tls.Config
 	if c.serverName == "" {
-		config, err = tlsconfigstore.GetTlsConfigurationForClient(serverName, cstream)
+		config, err = tlsconfigstore.GetTlsConfigurationForClient(serverName, cstream, token)
```
matthewstevenson88, line 86: I think the model should be: pass the token manager pointer to whatever API needs the token (so GetTlsConfigurationForClient, GetTlsConfigurationForServer), if that API has a local identity then get the token for that local identity and if not use the default, and then attach the token to the request.

internal/v2/BUILD
```diff
@@ -12,6 +12,7 @@ go_library(
   ],
   importpath = "github.com/google/s2a-go/internal/v2",
   deps = [
+    "//internal/tokenmanager",
```
matthewstevenson88, line 15: Please keep the deps alphabetized. :)

internal/v2/cert_verifier/cert_verifier_test.go
```diff
@@ -65,6 +67,16 @@ func TestVerifyClientCertChain(t *testing.T) {
 		log.Fatalf("error starting fake S2Av2 Server: %v", err)
 	}
 
+	os.Setenv("S2A_ACCESS_TOKEN", "token")
+	accessTokenManager, err := tokenmanager.NewSingleTokenAccessTokenManager()
+	if err != nil {
+		log.Fatalf("error generating token to authenticate to s2a: %v", err)
```
matthewstevenson88, line 73: nit: I think this should be "t.Fatalf" here, and similarly throughout the tests. Sorry if I missed this in previous reviews.

internal/v2/cert_verifier/cert_verifier_test.go
```diff
@@ -65,6 +67,16 @@ func TestVerifyClientCertChain(t *testing.T) {
 		log.Fatalf("error starting fake S2Av2 Server: %v", err)
 	}
 
+	os.Setenv("S2A_ACCESS_TOKEN", "token")
```
matthewstevenson88, line 70: In each unit test, please use a different token value (for example, in this test it could be "TestVerifyClientCertChain_s2a_access_token"). This helps prevent any dependence between the tests and make sure we reset everything correctly between each test.

internal/v2/BUILD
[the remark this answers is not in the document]
rmehta19: Will do! I'll fix the deps in this PR, but after this batch from last week are merged, I'll come back and do a cleanup PR to fix any remaining deps elsewhere in the v2 part of the repo.

internal/v2/cert_verifier/cert_verifier_test.go
rmehta19: Will do! Changed this from log to t in all three test suites

internal/v2/cert_verifier/cert_verifier.go
```diff
@@ -5,20 +5,25 @@ import (
...
 	s2av2pb "github.com/google/s2a-go/internal/proto/v2/s2a_go_proto"
 )
 
 // VerifyClientCertificateChain builds a SessionReq, sends it to S2Av2 and
 // receives a SessionResp.
-func VerifyClientCertificateChain(cstream s2av2pb.S2AService_SetUpSessionClient) func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
+func VerifyClientCertificateChain(cstream s2av2pb.S2AService_SetUpSessionClient, tokenManager tokenmanager.AccessTokenManager) func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
 	return func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
+		// Get Access Token
+		token, err := tokenManager.DefaultToken()
```
matthewstevenson88, line 18: Super sorry I missed this on the first pass, I forgot that the plan had changed here. We don't need to attach an auth mechanism in the cert verifier or the remote signer, only in the TLS config store.

internal/v2/s2av2_e2e_test.go
```diff
@@ -126,7 +129,7 @@ func runClient(ctx context.Context, t *testing.T, clientS2AAddress, serverAddr s
 }
 
 func TestEndToEndUsingFakeS2AOverTCP(t *testing.T) {
-	// TODO(rmehta19): Set S2A_ACCESS_TOKEN env variable once #46 merges.
+	os.Setenv(accessTokenEnvVariable, "")
```
matthewstevenson88, line 132: Why are we setting the env variable to the empty string here? Is the idea that we should have a test where there is no token? If so, I agree, but I think we should call `os.Unsetenv(accessTokenEnvVariable)` instead.

internal/v2/s2av2_test.go
```diff
@@ -102,17 +108,46 @@ func TestCloneClient(t *testing.T) {
 	if !ok {
 		t.Fatal("the created clone creds is not of type s2aTransportCreds")
 	}
-	if got, want := cmp.Equal(s2av2Creds, s2av2CloneCreds, protocmp.Transform(), cmp.AllowUnexported(s2av2TransportCreds{})), true; got != want {
-		t.Errorf("cmp.Equal(%v, %v) = %v, want %v", s2av2Creds, s2av2CloneCreds, got, want)
+	if got, want := cmp.Equal(s2av2Creds, s2av2CloneCreds, protocmp.Transform(), cmp.AllowUnexported(s2av2TransportCreds{}), cmp.Comparer(func(x, y tokenmanager.AccessTokenManager) bool {
```
matthewstevenson88, line 111: Nicely done!!

internal/v2/tls_config_store/tls_config_store.go
```diff
@@ -29,13 +30,16 @@ var (
 )
 
 // GetTlsConfigurationForClient returns a tls.Config instance for use by a client application.
-func GetTlsConfigurationForClient(serverHostname string, cstream s2av2pb.S2AService_SetUpSessionClient) (*tls.Config, error) {
+func GetTlsConfigurationForClient(serverHostname string, cstream s2av2pb.S2AService_SetUpSessionClient, tokenManager tokenmanager.AccessTokenManager) (*tls.Config, error) {
+	token, err := tokenManager.DefaultToken()
```
matthewstevenson88, line 34: This logic needs to be a little different: we need to accept a local identity pointer to GetTlsConfigurationForClient as well, if this pointer is non-nil then we call the tokenManager's Token(local identity) API, and if this pointer is nil then we call tokenManager.DefaultToken.

internal/v2/tls_config_store/tls_config_store.go
```diff
@@ -108,14 +112,17 @@ func GetTlsConfigurationForClient(serverHostname string, cstream s2av2pb.S2AServ
 }
 
 // GetTlsConfigurationForServer returns a tls.Config instance for use by a server application.
-func GetTlsConfigurationForServer(cstream s2av2pb.S2AService_SetUpSessionClient) (*tls.Config, error) {
+func GetTlsConfigurationForServer(cstream s2av2pb.S2AService_SetUpSessionClient, tokenManager tokenmanager.AccessTokenManager) (*tls.Config, error) {
```
matthewstevenson88, line 115: The logic on the server-side is a little more complicated (because of the SNI extension), but for now we should do the following: we need to accept an array of local identity pointers to GetTlsConfigurationForServer as well, and iterate through the list, for each pointer, if this pointer is non-nil then we call the tokenManager's Token(local identity) API, and if this pointer is nil then we call tokenManager.DefaultToken.

When we move the call to S2Av2 to a helper function in the follow-up PR, we will also need to add an extra auth mechanism for the SNI. Please add a TODO so we don't forget to do this. :)

internal/v2/cert_verifier/cert_verifier.go
rmehta19: No worries, Thanks for letting me know! This is because S2A only looks at the token attached to the first SessionReq sent on the stream right, and TLS config store is always going to make the first SessionReq?

internal/v2/s2av2_e2e_test.go
rmehta19: I believe I set it to an empty string to replicate the structure of the test in the v1 implementation. I can set a non empty string token here, and create another test with no token using Unsetenv, and make sure the error is being caught.

internal/v2/s2av2_test.go
rmehta19: Thanks!

internal/v2/tls_config_store/tls_config_store.go
rmehta19: Thanks for the explanation! A question: what if there are > 1 non - nil local identities in the array? How do we know we've picked the right identity to get the corresponding token of? 

I'm assuming that once we integrate the SNI extension, we can use the SNI to pick with local identity from the list we would like to get the corresponding token for, but for now, will the array of local identities just be of size 1?

rmehta19:
Thanks for the comments Matt!
Note: I'll squash down the commits to 1 before merging this PR.

internal/v2/tls_config_store/tls_config_store.go
matthewstevenson88: > A question: what if there are > 1 non - nil local identities in the array? How do we know we've picked the right identity to get the corresponding token of?

We will include one AuthenticationMechanism for each in the array and the S2A will select which one to use.

> I'm assuming that once we integrate the SNI extension, we can use the SNI to pick with local identity from the list we would like to get the corresponding token for, but for now, will the array of local identities just be of size 1?

When we have the SNI, we will add one extra AuthenticationMechanism that has the SNI local identity and the corresponding token.

For now, yes the array will be of size 1.

internal/v2/tls_config_store/tls_config_store.go
```diff
@@ -29,13 +30,25 @@ var (
 )
 
 // GetTlsConfigurationForClient returns a tls.Config instance for use by a client application.
-func GetTlsConfigurationForClient(serverHostname string, cstream s2av2pb.S2AService_SetUpSessionClient) (*tls.Config, error) {
+func GetTlsConfigurationForClient(serverHostname string, cstream s2av2pb.S2AService_SetUpSessionClient, tokenManager tokenmanager.AccessTokenManager, localIdentity *commonpbv1.Identity) (*tls.Config, error) {
+	var token string
```
matthewstevenson88, line 34: Could we extract the logic for how to produce the AuthenticationMechanism array into a helper function that can be used in both GetTlsConfigurationForClient and GetTlsConfigurationForServer?

We can model it on the v1 code here: https://github.com/google/s2a-go/blob/8249b7935631c5db8ad4d77e1d15a965e14c7d48/internal/handshaker/handshaker.go#L397

internal/v2/tls_config_store/tls_config_store.go
```diff
@@ -199,6 +194,55 @@ func GetTlsConfigurationForServer(cstream s2av2pb.S2AService_SetUpSessionClient)
...
+				Identity: localIdentity,
+				MechanismOneof: &s2av2pb.AuthenticationMechanism_Token {
+					Token: token,
+				},
+			})
+		} else {
+			token, err := tokenManager.Token(localIdentity)
+			if err != nil {
+				grpclog.Infof("unable to get token for local identity %v: %v", localIdentity, err)
+				return nil
```
matthewstevenson88, line 233: Should we "continue" here rather than returning?

matthewstevenson88 (approved):

internal/v2/tls_config_store/tls_config_store.go
rmehta19: Yes, thanks for catching this!

PR #47, opened 2022-06-23
Add GetConfigForClient field to Server TLS Config

rmehta19:
Add GetConfigForClient field to Server TLS Config. 

This allows the Server to use SNI from ClientHello to return a config with the certificate that the client expects, and establish a secure connection. 

Added corresponding unit tests to TLS Config store test suite. While adding these tests, cleaned up existing tests to ensure stream creation happens during each unit tests(in future if we add failing tests, stream needs to be recreated; it cannot be reused by failing / passing tests)

### DIFF
--- a/internal/v2/tls_config_store/tls_config_store.go
+++ b/internal/v2/tls_config_store/tls_config_store.go
@@ -107,8 +107,89 @@ func GetTlsConfigurationForClient(serverHostname string, cstream s2av2pb.S2AServ
 
 // GetTlsConfigurationForServer returns a tls.Config instance for use by a server application.
 func GetTlsConfigurationForServer(cstream s2av2pb.S2AService_SetUpSessionClient, tokenManager tokenmanager.AccessTokenManager, localIdentities []*commonpbv1.Identity) (*tls.Config, error) {
-	// TODO(rmehta19): Add an extra auth mechanism for the SNI.
-	// TODO(rmehta19): move call to S2Av2 to a helper function.
+	return &tls.Config {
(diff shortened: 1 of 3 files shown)


internal/v2/tls_config_store/tls_config_store.go
```diff
@@ -107,8 +107,89 @@ func GetTlsConfigurationForClient(serverHostname string, cstream s2av2pb.S2AServ
...
+			if err != nil {
+				return nil, err
+			}
+			cert.Certificate = append(cert.Certificate, x509Cert.Raw)
+			if i == 0 {
+				cert.Leaf = x509Cert
+			}
+		}
+
+		cert.PrivateKey = remotesigner.New(cert.Leaf, cstream, &commonpbv1.Identity {
```
matthewstevenson88, line 149: Now that we only send auth mechanisms in the GetTlsConfigurationRequest, should we remove the local identity from the remote signer? IIUC, it is no longer needed.

matthewstevenson88 (commented):
Generally looks good, just one comment/question about something left over in the remote signer API.

internal/v2/tls_config_store/tls_config_store.go
rmehta19: In remote signer, I use the localIdentity field of the SessionReq to determine who is requesting the signing operation, a client using s2a or a server using s2a. This field is used on the fakes2av2 side to determine which private key to use to perform the signing. 

Perhaps in the next PR, when I add the token managing feature to fakes2av2, I can use that field to determine who is using s2av2.

PR #55, opened 2022-06-30
Manage token in Fake S2Av2

rmehta19:
Fake S2Av2 checks Authentication Mechanisms attached to the first ```SessionReq```(first ```SessionReq``` on the client stream always contains a ```GetTlsConfigurationReq```) sent on the client stream. Fake S2Av2 verifies that at least one of these Authentication Mechanisms has a valid token. Fake S2Av2 stores a valid token: ```valid_token``` in Fake S2Av2 ```Server``` struct. If there is no valid token attached to the first ```SessionReq```, S2Av2 closes the stream.

Modify Remote Signer API to not accept ```Identity```. Add ```isClientSide``` field to Fake S2Av2 ```Server``` struct, which is used when a ```OffloadPrivateKeyOperationReq``` ```SessionReq``` is sent on the client stream. Populate ```isClientSide``` according to ```ConnectionSide``` field in ```GetTlsConfigurationReq``` when ```GetTlsConfigurationReq``` ```SessionReq``` is sent on the client stream. Note that ```GetTlsConfigurationReq``` always the first type of ```SessionReq``` sent on the stream.

Add additional / modify existing unit tests for Remote Signer and TLS Configuration Store library to test changes.


internal/v2/fakes2av2/fakes2av2.go
```diff
@@ -34,11 +34,17 @@ var (
 // Server is a fake S2A Server for testing.
 type Server struct {
 	s2av2pb.UnimplementedS2AServiceServer
+	validToken string
+	isClientSide bool
```
matthewstevenson88, line 38: nit: could we call this variable "isAssistingClientSide" instead?

Rationale: the fake S2A assists a client or server, but the fake S2A itself is always acting as a server.

internal/v2/fakes2av2/fakes2av2.go
```diff
@@ -34,11 +34,17 @@ var (
...
+	// TODO(rmehta19): Decide whether to also store validationResult (bool).
```
matthewstevenson88, line 39: For my own understanding: does "validationResult" mean "yes at least one auth mech was successfully validated vs. no none of the auth mechs were successfully validated"?

internal/v2/fakes2av2/fakes2av2.go
```diff
@@ -79,31 +98,44 @@ func (s *Server) SetUpSession(stream s2av2pb.S2AService_SetUpSessionServer) erro
...
+		s.isClientSide = true
+	case commonpb.ConnectionSide_CONNECTION_SIDE_SERVER:
+		s.isClientSide = false
+	default:
+		return fmt.Errorf("unknown ConnectionSide, req.GetGetTlsConfigurationReq().GetConnectionSide() returned %v", connSide)
+	}
+	return nil
+}
+
+func (s *Server) isValidToken(authMechanisms []*s2av2pb.AuthenticationMechanism) error {
```
matthewstevenson88, line 113: nit: could we call this method "hasValidToken" instead?

internal/v2/fakes2av2/fakes2av2.go
```diff
@@ -34,11 +34,17 @@ var (
...
+	isClientSide bool
+	// TODO(rmehta19): Decide whether to also store validationResult (bool).
+	// Set this after validating token attached to first SessionReq. Check
+	// this field before completing subsequent SessionReq.
 }
 
 // SetUpSession receives SessionReq, performs request, and returns a
 // SessionResp, all on the server stream.
 func (s *Server) SetUpSession(stream s2av2pb.S2AService_SetUpSessionServer) error {
+	s.validToken = "valid_token"
```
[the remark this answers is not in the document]
rmehta19: Thanks for pointing this out, will do!

internal/v2/fakes2av2/fakes2av2.go
rmehta19: I meant for validationResult to indicate at least one auth mechanism was successfully validated. If I added this field, it would enforce that SessionReq with ```GetTlsConfigurationReq``` needs to be sent on the stream before anything else. This would mainly just affect the unit tests I have for the client libraries. Before I send a ```ValidatePeerCertificateChainReq```, I would need to send a SessionReq with a ```GetTlsConfigurationReq```

matthewstevenson88: Understood, this plan SGTM.

internal/v2/fakes2av2/fakes2av2.go
```diff
@@ -34,6 +34,11 @@ var (
 // Server is a fake S2A Server for testing.
 type Server struct {
 	s2av2pb.UnimplementedS2AServiceServer
+	ExpectedToken string
```
matthewstevenson88, line 37: Please add a comment for the exported field. :)

internal/v2/s2av2_e2e_test.go
```diff
@@ -19,7 +19,7 @@ import (
 
 const (
 	accessTokenEnvVariable = "S2A_ACCESS_TOKEN"
-	testAccessToken        = "test_access_token"
+	testAccessToken        = "valid_token"
```
matthewstevenson88, line 22: Is this variable still needed?

matthewstevenson88 (approved):
LGTM!

internal/v2/s2av2_e2e_test.go
rmehta19: No, it's not, since I'm setting a different token in each Test, thanks for catching this!

rmehta19:
Thanks for the review!

PR #57, opened 2022-07-07
Populate localIdentity in first SessionReq.

rmehta19:
Populate LocalIdentity field in first SessionReq to S2A (from TLS Config Store)

### DIFF
--- a/internal/v2/s2av2_e2e_test.go
+++ b/internal/v2/s2av2_e2e_test.go
@@ -10,6 +10,7 @@ import (
 	"testing"
 	"time"
 
+	commonpbv1 "github.com/google/s2a-go/internal/proto/common_go_proto"
 	helloworldpb "github.com/google/s2a-go/internal/proto/examples/helloworld_go_proto"
 	s2av2pb "github.com/google/s2a-go/internal/proto/v2/s2a_go_proto"
 	"github.com/google/s2a-go/internal/v2/fakes2av2"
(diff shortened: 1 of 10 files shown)


internal/v2/s2av2_e2e_test.go
```diff
@@ -154,11 +165,21 @@ func TestEndToEndUsingFakeS2AOnUDS(t *testing.T) {
 	grpclog.Infof("fake S2A for client listening on UDS at address: %v", clientS2AAddr)
 
 	// Start the server.
-	serverAddr := startServer(t, serverS2AAddr)
+	var localIdentities []*commonpbv1.Identity
```
matthewstevenson88, line 168: For each unit test, can we actually have 2 unit tests: one with no local identity set, one with the local identity set?

internal/v2/tls_config_store/tls_config_store.go
```diff
@@ -33,8 +33,14 @@ var (
 // GetTlsConfigurationForClient returns a tls.Config instance for use by a client application.
 func GetTlsConfigurationForClient(serverHostname string, cstream s2av2pb.S2AService_SetUpSessionClient, tokenManager tokenmanager.AccessTokenManager, localIdentity *commonpbv1.Identity) (*tls.Config, error) {
 	authMechanisms := getAuthMechanisms(tokenManager, []*commonpbv1.Identity{localIdentity})
+	var locId *commonpbv1.Identity
```
matthewstevenson88, line 36: Do we need to have the `locId` temporary variable? Can we directly use the `localIdentity` argument?

matthewstevenson88 (commented):
Thanks for the PR Riya! Generally LGTM, just 2 minor comments. :)

internal/v2/s2av2_e2e_test.go
rmehta19: Will do! I separated all the E2E tests, so that in one, both the client and server set localIdentity, localIdentities, and in the other, neither client or server set localIdentity, localIdentities.

PR #59, opened 2022-07-08
Add VerificationMode to ValidatePeerCertificateChainReq SessionReq

rmehta19:
Add VerificationMode to ValidatePeerCertificateChainReq SessionReq.

### DIFF
--- a/internal/v2/s2av2.go
+++ b/internal/v2/s2av2.go
@@ -33,12 +33,13 @@ type s2av2TransportCreds struct {
 	// localIdentity should only be used by the client.
 	localIdentity *commonpbv1.Identity
 	// localIdentities should only be used by the server.
-	localIdentities []*commonpbv1.Identity
+	localIdentities  []*commonpbv1.Identity
+	verificationMode s2av2pb.ValidatePeerCertificateChainReq_VerificationMode
 }
 
 // NewClientCreds returns a client-side transport credentials object that uses
 // the S2Av2 to establish a secure connection with a server.
-func NewClientCreds(s2av2Address string, localIdentity *commonpbv1.Identity) (credentials.TransportCredentials, error) {
+func NewClientCreds(s2av2Address string, localIdentity *commonpbv1.Identity, verificationMode s2av2pb.ValidatePeerCertificateChainReq_VerificationMode) (credentials.TransportCredentials, error) {
 	// Create an AccessTokenManager instance to use to authenticate to S2Av2.
 	accessTokenManager, err := tokenmanager.NewSingleTokenAccessTokenManager()
 	if err != nil {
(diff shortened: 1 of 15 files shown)


internal/v2/s2av2.go
```diff
@@ -29,11 +29,12 @@ type s2av2TransportCreds struct {
 	serverName string
 	s2av2Address string
 	tokenManager *tokenmanager.AccessTokenManager
+	verificationMode s2av2pb.ValidatePeerCertificateChainReq_VerificationMode
 }
 
 // NewClientCreds returns a client-side transport credentials object that uses
 // the S2Av2 to establish a secure connection with a server.
-func NewClientCreds(s2av2Address string) (credentials.TransportCredentials, error) {
+func NewClientCreds(s2av2Address string, verificationMode s2av2pb.ValidatePeerCertificateChainReq_VerificationMode) (credentials.TransportCredentials, error) {
```
matthewstevenson88, line 37: We don't want to expose the proto directly to the users, so I think the right way to do this is:

1. Add a "verification mode" enum to the options file: https://github.com/google/s2a-go/blob/main/s2a_options.go

2. Add a "verification mode" field to each of the client/server options, which defaults to the "connecting to Google" value

3. Plumb this value through to cert_verifier.go, and then convert to the proto enum when preparing the SessionReq message.

Does that sound reasonable?

rmehta19: Thanks for pointing this out! In this PR, I added a verificationMode field to the client and server options in s2a_options, but I did expose the proto, I'll change it to the enum.

s2a_options.go
```diff
@@ -64,6 +64,14 @@ func NewUID(name string) Identity {
 	return &uid{uid: name}
 }
 
+type VerificationModeType int
```
matthewstevenson88, line 67: Please add a top-level comment for this, e.g. "VerificationModeType specifies the mode that S2A must use to verify the peer certificate chain."

s2a_options.go
```diff
@@ -105,11 +113,16 @@ type ClientOptions struct {
 	EnsureProcessSessionTickets *sync.WaitGroup
 	// If true, enables the use of S2Av2.
 	EnableV2 bool
+	// VerificationMode specifies the verification mode supported by S2Av2.
```
matthewstevenson88, line 116: Suggested edit: VerificationMode specifies the mode that S2A must use to verify the peer certificate chain.

s2a_options.go
```diff
@@ -123,11 +136,16 @@ type ServerOptions struct {
 	S2AAddress string
 	// If true, enables the use of S2Av2.
 	EnableV2 bool
+	// VerificationMode specifies the verification mode supported by S2Av2.
```
matthewstevenson88, line 139: Suggested edit: VerificationMode specifies the mode that S2A must use to verify the peer certificate chain.

matthewstevenson88 (approved):
LGTM with 3 nits.

s2a_options.go
rmehta19: Will do, thanks!

s2a_options.go
rmehta19: Will do, thanks!

s2a_options.go
rmehta19: Will do, thanks!

PR #65, opened 2022-07-14
Fix Lint Errors

rmehta19:
Fix lint errors.

### DIFF
--- a/internal/v2/fakes2av2_server/fakes2av2_server.go
+++ b/internal/v2/fakes2av2_server/fakes2av2_server.go
@@ -1,3 +1,4 @@
+// Package main runs an S2Av2 service.
 package main
 
 import (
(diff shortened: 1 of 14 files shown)


internal/v2/fakes2av2_server/fakes2av2_server.go
```diff
@@ -1,3 +1,4 @@
+// Package fakes2av2_server runs an S2Av2 service.
```
matthewstevenson88, line 1: Should this be "package main"?

matthewstevenson88 (approved):

internal/v2/fakes2av2_server/fakes2av2_server.go
rmehta19: Yes! Thanks for catching this.

PR #72, opened 2022-07-25
Add NextProtos field to TLS config.

rmehta19:


internal/v2/tls_config_store/tls_config_store.go
```diff
@@ -124,6 +124,7 @@ func GetTLSConfigurationForClient(serverHostname string, cstream s2av2pb.S2AServ
 		ClientSessionCache:    nil,
 		MinVersion:            minVersion,
 		MaxVersion:            maxVersion,
+		NextProtos:            []string{"h2"},
```
matthewstevenson88, line 127: nit: can we make "h2" a constant in this file?

matthewstevenson88, line 127: Please update the unit tests to check that NextProtos is what we expect.

rmehta19: Will do!

matthewstevenson88 (approved):

matthewstevenson88 (approved):

PR #75, opened 2022-08-09
Update Cipher Suites in proto and populate in Server TLS Config.

rmehta19:
When negotiating TLS 1.2, populate the CipherSuites field in the server side TLS config.

Run ./tools/proto/regenerate_proto.sh to update protocol buffer generated code.


rmehta19:
I am also thinking about adding logic(in a follow up PR) to the fake S2Av2 service layer to return TLS 1.2 in the Min / Max TLS Version so that we can update the E2E tests/ tls config store unit tests in this open source repo to test TLS 1.2 with this cipher suite change. Do you think this is needed?

internal/proto/v2/common.proto
```diff
@@ -18,26 +18,40 @@ package s2a.proto.v2;
 
 option go_package = "github.com/google/s2a/internal/proto/v2/common_go_proto";
 
-// The ciphersuites for the TLS handshake supported by S2A. Note that handshake
-// ciphersuites are NOT configurable in TLS 1.3.
+// DEPRECATED: The ciphersuites for the TLS handshake supported by S2A. Note
+// that handshake ciphersuites are NOT configurable in TLS 1.3.
+// TODO: Delete this enum when it is no longer in use.
```
matthewstevenson88, line 23: Please assign all TODOs in the protos to my Github username.

internal/v2/tls_config_store/tls_config_store.go
```diff
@@ -186,20 +184,54 @@ func ClientConfig(tokenManager tokenmanager.AccessTokenManager, localIdentities
 
 		clientAuth := getTLSClientAuthType(tlsConfig)
 
+		var cipherSuites []uint16
+		if maxVersion <= tls.VersionTLS12 {
```
matthewstevenson88, line 188: nit: please omit the TLS version check, just to future proof against e.g. configurable ciphersuites in TLS 1.3.

matthewstevenson88:
> I am also thinking about adding logic(in a follow up PR) to the fake S2Av2 service layer to return TLS 1.2 in the Min / Max TLS Version so that we can update the E2E tests/ tls config store unit tests in this open source repo to test TLS 1.2 with this cipher suite change. Do you think this is needed?

Ya that would be useful if you have time! Thank you!

matthewstevenson88 (approved):

rmehta19:
> > I am also thinking about adding logic(in a follow up PR) to the fake S2Av2 service layer to return TLS 1.2 in the Min / Max TLS Version so that we can update the E2E tests/ tls config store unit tests in this open source repo to test TLS 1.2 with this cipher suite change. Do you think this is needed?
> 
> Ya that would be useful if you have time! Thank you! As you suggest, we can do that in a follow-up PR.

Will do! I'll add a TODO so we don't miss this.

internal/proto/v2/common.proto
rmehta19: Done! I also reran the regenerate_proto.sh script to sync the protos and the generated code.

internal/v2/tls_config_store/tls_config_store.go
rmehta19: Good point! Done!

PR #80, opened 2022-08-29
Allow GetTlsConfiguration RPC to return empty credentials to a client.

matthewstevenson88:
This is needed to support TLS (rather than mTLS).

### DIFF
--- a/internal/v2/fakes2av2/fakes2av2.go
+++ b/internal/v2/fakes2av2/fakes2av2.go
@@ -58,8 +58,11 @@ var (
 type Server struct {
 	s2av2pb.UnimplementedS2AServiceServer
 	// ExpectedToken is the token S2Av2 expects to be attached to the SessionReq.
-	ExpectedToken         string
-	isAssistingClientSide bool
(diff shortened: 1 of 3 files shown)


internal/v2/fakes2av2/fakes2av2.go
```diff
@@ -213,8 +216,25 @@ func validatePeerCertificateChain(req *s2av2pb.ValidatePeerCertificateChainReq)
 }
 
 // TODO(rmehta19): Update this to return ciphersuites in Client/Server TlsConfiguration.
```
tavishvaidya, line 218: Should this comment be deleted?

tavishvaidya (approved):

internal/v2/fakes2av2/fakes2av2.go
matthewstevenson88: I think so yes, good catch. I'm going to hold off doing so here because I should do a careful pass of Riya's TODOs and see what (if anything) needs to be done. Will send you a follow-up PR for this.

PR #84, opened 2023-01-13
expose public method to generate a s2a-enabled tls configuration

xmenxk:
bazel build ...
bazel test ...

### DIFF
--- a/s2a.go
+++ b/s2a.go
@@ -22,8 +22,10 @@ package s2a
 
 import (
 	"context"
+	"crypto/tls"
 	"errors"
 	"fmt"
+	"github.com/google/s2a-go/internal/tokenmanager"
 	"net"
 	"sync"
 	"time"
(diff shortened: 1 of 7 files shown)


s2a.go
```diff
@@ -276,6 +278,52 @@ func (c *s2aTransportCreds) OverrideServerName(serverNameOverride string) error
...
+// TLSClientConfigOptions specifies parameters for creating client TLS config.
+type TLSClientConfigOptions struct{}
+
+// TLSClientConfigFactory defines the interface for a client TLS config factory.
+type TLSClientConfigFactory interface {
+	Build(ctx context.Context, opts *TLSClientConfigOptions) (*tls.Config, error)
+}
+
+// NewTLSClientConfigFactory returns an instance of s2aTLSClientConfigFactory.
+func NewTLSClientConfigFactory(opts *ClientOptions) (TLSClientConfigFactory, error) {
```
matthewstevenson88, line 290: Can we add tests for this API, similar to what you did in s2av2_e2e_test.go?

matthewstevenson88 (commented):
Thanks for the PR Kui! Looks good, just a request for some more tests. :)

matthewstevenson88:
Also it looks like one of the presubmits failed - can you fix whatever the test that this PR breaks?

s2a.go
xmenxk: done, thanks!

xmenxk:
> Also it looks like one of the presubmits failed - can you fix whatever the test that this PR breaks?

seems like a transient issue, all passing now, PTAL :)

matthewstevenson88 (approved):
Thanks for the PR Kui!

matthewstevenson88:
Please "squash and merge" rather than just merging.

PR #86, opened 2023-01-17
Add support for AppEngine-specific dialer.

matthewstevenson88:
The default gRPC dialer does not seem to work on AppEngine. This will be e2e tested by deploying a Go app to AppEngine.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -55,6 +55,16 @@ go_repository(
   version = "v1.45.0",
 )
 
+# Google AppEngine.
+#
+# Last Updated: January 17, 2023.
+go_repository(
+  name = "org_golang_google_appengine",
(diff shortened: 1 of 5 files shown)


tavishvaidya (commented):
Are there any unit tests that needs to be updated?

tavishvaidya (approved):
Are there any unit tests that will need to be updated?

tavishvaidya (approved):

matthewstevenson88:
Unfortunately we can't easily trigger the new codepath in a unit test, since it is gated by an API that checks whether the application is running on AppEngine. For safety, I've flag-guarded this feature and we will prioritize setting up a Kokoro test that launches an AppEngine app and tests out this code.

PR #89, opened 2023-02-06
add an option to pass server name into s2a's tls config

xmenxk:
during handshake, s2a will check if the hostname found in server's certificate matches the expected server name. Currently s2a does not allow specifying this expected server name and using an empty string, which leads to error in handshake.

this PR makes it possible to specify such server name when building a client tls config instance.

bazel test ...
bazel build ...

### DIFF
--- a/internal/v2/s2av2.go
+++ b/internal/v2/s2av2.go
@@ -114,10 +114,7 @@ func (c *s2av2TransportCreds) ClientHandshake(ctx context.Context, serverAuthori
 		return nil, nil, errors.New("client handshake called using server transport credentials")
 	}
 	// Remove the port from serverAuthority.
-	serverName, _, err := net.SplitHostPort(serverAuthority)
-	if err != nil {
-		serverName = serverAuthority
-	}
+	serverName := removeServerNamePort(serverAuthority)
 	ctx, cancel := context.WithTimeout(ctx, defaultTimeout)
 	defer cancel()
 	cstream, err := createStream(ctx, c.s2av2Address)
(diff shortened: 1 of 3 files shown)


internal/v2/s2av2_e2e_test.go
```diff
@@ -268,7 +268,7 @@ func TestNewClientTlsConfigWithTokenManager(t *testing.T) {
 	}
 	ctx, cancel := context.WithTimeout(context.Background(), defaultE2ETimeout)
 	defer cancel()
-	config, err := NewClientTLSConfig(ctx, s2AAddr, accessTokenManager, s2av2pb.ValidatePeerCertificateChainReq_CONNECT_TO_GOOGLE)
+	config, err := NewClientTLSConfig(ctx, s2AAddr, accessTokenManager, s2av2pb.ValidatePeerCertificateChainReq_CONNECT_TO_GOOGLE, "test")
```
matthewstevenson88, line 271: Please use a more explanatory variable name than "test". :)

internal/v2/s2av2_e2e_test.go
```diff
@@ -288,7 +288,7 @@ func TestNewClientTlsConfigWithoutTokenManager(t *testing.T) {
 	var tokenManager tokenmanager.AccessTokenManager
 	ctx, cancel := context.WithTimeout(context.Background(), defaultE2ETimeout)
 	defer cancel()
-	config, err := NewClientTLSConfig(ctx, s2AAddr, tokenManager, s2av2pb.ValidatePeerCertificateChainReq_CONNECT_TO_GOOGLE)
+	config, err := NewClientTLSConfig(ctx, s2AAddr, tokenManager, s2av2pb.ValidatePeerCertificateChainReq_CONNECT_TO_GOOGLE, "test")
```
matthewstevenson88, line 291: Same here.

s2a.go
```diff
@@ -279,7 +279,9 @@ func (c *s2aTransportCreds) OverrideServerName(serverNameOverride string) error
 }
 
 // TLSClientConfigOptions specifies parameters for creating client TLS config.
-type TLSClientConfigOptions struct{}
+type TLSClientConfigOptions struct {
+	ServerName string
```
matthewstevenson88, line 283: Can you clarify how we expect the user to populate this value in the HTTP case?

matthewstevenson88, line 283: Please add a comment description for this field.

matthewstevenson88 (commented):
Thanks for the PR Kui! Please clean up the title so it is a bit more descriptive and also add some background info in the PR on why this is necessary. :)

s2a.go
xmenxk: here's an example, I also added it to the comment description for the field
		tlsConfig, err := factory.Build(ctx, &s2a.TLSClientConfigOptions{
			ServerName: "example.com",
		})

matthewstevenson88: Sorry let me clarify: in the GUAC HTTP code, how do you know what value you should populate the `ServerName` field with?

internal/v2/s2av2.go
```diff
@@ -244,28 +241,35 @@ func NewClientTLSConfig(
...
-		serverName = serverNameOverride
-	}
+	serverName := removeServerNamePort(serverNameOverride)
 	c.info.ServerName = serverName
 	c.serverName = serverName
 	return nil
 }
 
+// Remove the trailing port from server name.
+func removeServerNamePort(serverName string) string {
```
matthewstevenson88, line 265: Can we add unit tests for this API?

matthewstevenson88 (approved):

PR #90, opened 2023-02-23
add optional fallback capability

xmenxk:
changes in this PR:
1. for grpc: add an optional fallback address to s2aV2 transport credentials, so that when dialing with s2a fails, it will try regular TLS with the fallback addr, if specified.
2. for http: add a dialtls function which can dial mtls with s2aV2, an optional fallback addr can be provided as well.


matthewstevenson88 (commented):
Apologies in advance for the back and forth - I don't think I understood the scope of the proposed change when we discussed offline. I'm concerned about this PR for 2 reasons:

1. It has nontrivial functional changes for *all* S2A-Go users, regardless of whether they want a fallback or not (e.g. closing the connection on failures, when traditionally this is the responsibility of gRPC).
2. It prescribes a very specific fallback mechanism (which is baking a specific use case of these libraries into the code).

I think this needs to be re-architected a bit so that (a) if the user does not explicitly request a fallback, then there are no functional changes, and (b) we allow the user to inject the specifics of the fallback mechanism, rather than prescribing how it is done.

WDYT?

To minimize back-and-forth, it may be worthwhile to share the proposed changes to the top-level API in a doc first.

internal/v2/s2av2.go
```diff
@@ -59,18 +63,25 @@ type s2av2TransportCreds struct {
 
 // NewClientCreds returns a client-side transport credentials object that uses
 // the S2Av2 to establish a secure connection with a server.
-func NewClientCreds(s2av2Address string, localIdentity *commonpbv1.Identity, verificationMode s2av2pb.ValidatePeerCertificateChainReq_VerificationMode) (credentials.TransportCredentials, error) {
+func NewClientCreds(fallbackServerAddr string, s2av2Address string, localIdentity *commonpbv1.Identity, verificationMode s2av2pb.ValidatePeerCertificateChainReq_VerificationMode) (credentials.TransportCredentials, error) {
```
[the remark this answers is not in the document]
xmenxk: thx Matt,
This is our internal function, user specifies the fallback addr via ClientOptions, where I put the explanation, plz see s2a_options.go
The public interface is e.g.:

transportCreds, _ = s2a.NewClientCreds(&s2a.ClientOptions{
					S2AAddress:         s2aAddress,
					EnableV2:           true,
					FallbackServerAddr: o.DefaultEndpoint,
				})

which will then call our internal function and pass in the fallback addr.

.gitignore
```diff
@@ -4,3 +4,5 @@ bazel-*
 //example/client/client
 //example/server/server
 //internal/v2/fakes2av2_server/fakes2av2_server
+
+.idea/
```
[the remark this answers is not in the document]
xmenxk: it's a directory that intelliJ IDE creates

xmenxk:
> To minimize back-and-forth, it may be worthwhile to share the proposed changes to the top-level API in a doc first.

SG, I'll put something in a doc and we can discuss there. thanks!

fallback/s2a_fallback.go
```diff
@@ -0,0 +1,121 @@
...
+	"google.golang.org/grpc/grpclog"
+	"net"
+)
+
+const (
+	alpnProtoStrH2   = "h2"
+	defaultHttpsPort = "443"
+)
+
+// DefaultFallbackClientHandshakeFunc returns an implementation of FallbackOptions.FallbackClientHandshakeFunc.
```
matthewstevenson88, line 17: Please explain what the default fallback behavior is. :)

fallback/s2a_fallback.go
```diff
@@ -0,0 +1,121 @@
...
+//	transportCreds, _ = s2a.NewClientCreds(&s2a.ClientOptions{
+//		S2AAddress: s2aAddress,
+//		EnableV2:   true,
+//		FallbackOpts: &s2a.FallbackOptions{ // optional, and for V2 only
+//			FallbackClientHandshakeFunc: fallback.DefaultFallbackClientHandshakeFunc(fallbackAddr),
+//		},
+//	})
+//
+// The fallbackAddr is expected to be a network address, e.g. example.com:port. If port is not specified,
+// it uses default port 443
```
matthewstevenson88, line 29: micro-nit: Missing period at the end of the sentence.

fallback/s2a_fallback.go
```diff
@@ -0,0 +1,121 @@
...
+func DefaultFallbackClientHandshakeFunc(fallbackAddr string) func(context.Context, string, net.Conn, error) (net.Conn, credentials.AuthInfo, error) {
+	return func(ctx context.Context, originServer string, originConn net.Conn, originErr error) (net.Conn, credentials.AuthInfo, error) {
```
matthewstevenson88, line 31: Can you clarify the meaning of the originServer and originConn and originErr parameters?

It's unclear to me what these mean.

fallback/s2a_fallback.go
```diff
@@ -0,0 +1,121 @@
...
+		fallbackServerAddr, fallbackServerName, err := processFallbackAddr(fallbackAddr)
+		if err != nil {
+			return nil, nil, fmt.Errorf("no fallback server address specified, skipping fallback; S2Av2 client handshake with %s error: %w", originServer, originErr)
```
matthewstevenson88, line 34: micro-nit: in the logs, please say "S2A" rather than "S2Av2". We want to ensure that the v1 vs. v2 difference is an internal implementation detail that is not visible to external customers.

fallback/s2a_fallback.go
```diff
@@ -0,0 +1,121 @@
...
+			grpclog.Infof("ConnectionState.NegotiatedProtocol: %v", tc.ConnectionState().NegotiatedProtocol)
+			grpclog.Infof("ConnectionState.HandshakeComplete: %v", tc.ConnectionState().HandshakeComplete)
+			grpclog.Infof("ConnectionState.ServerName: %v", tc.ConnectionState().ServerName)
+		}
+		originConn.Close()
+		return fbConn, tlsInfo, nil
+	}
+}
+
+// DefaultFallbackDialerAndAddress returns a TLS dialer and a network address for it to dial with.
```
matthewstevenson88, line 70: Same here, please explain the fallback functionality.

fallback/s2a_fallback.go
```diff
@@ -0,0 +1,121 @@
...
+//			FallbackOpts: &s2a.FallbackOptions{
+//				FallbackDialer: &s2a.FallbackDialer{
+//					Dialer:     fallbackDialer,
+//					ServerAddr: fallbackServerAddr,
+//				},
+//			},
+//	})
+//
+// The fallbackAddr is expected to be a network address, e.g. example.com:port. If port is not specified,
+// it uses default port 443
```
matthewstevenson88, line 86: micro-nit: Missing period.

matthewstevenson88 (commented):
The new code structure broadly LGTM, did a pass of s2a_fallback.go. Please let me know when the PR is ready for another pass. :)

fallback/s2a_fallback.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+func processFallbackAddr(fallbackAddr string) (string, string, error) {
+	// hostname:port
+	var fallbackServerAddr string
+	// hostname
+	var fallbackServerName string
+	var err error
+
+	if fallbackAddr != "" {
```
matthewstevenson88, line 128: nit: Can we reverse the logic here so that we first check if fallbackAddr is the empty string and if so return early, and then have the rest of the logic?

fallback/s2a_fallback_test.go
```diff
@@ -0,0 +1,101 @@
...
+			true,
+		},
+		{
+			"test empty fallback address",
+			"",
+			false,
+			false,
+		},
+	} {
+		fbFunc, err := DefaultFallbackClientHandshakeFunc(tc.inputFallbackAddr)
```
matthewstevenson88, line 51: Both here and below, we should be executing the fallback function or fallback dialer in the tests. :)

internal/v2/s2av2_test.go
```diff
@@ -49,7 +49,7 @@ func TestNewClientCreds(t *testing.T) {
 				IdentityOneof: &commonpbv1.Identity_Hostname{
 					Hostname: "test_rsa_client_identity",
 				},
-			}, s2av2pb.ValidatePeerCertificateChainReq_CONNECT_TO_GOOGLE)
+			}, s2av2pb.ValidatePeerCertificateChainReq_CONNECT_TO_GOOGLE, nil)
```
matthewstevenson88, line 52: Should one of the tests pass in a fallback and make sure it works as expected?

internal/v2/s2av2_e2e_test.go
```diff
@@ -130,7 +130,7 @@ func startServer(t *testing.T, s2aAddress string, localIdentities []*commonpbv1.
 
 // runClient starts up a client and calls the server.
 func runClient(ctx context.Context, t *testing.T, clientS2AAddress, serverAddr string, localIdentity *commonpbv1.Identity) {
-	creds, err := NewClientCreds(clientS2AAddress, localIdentity, s2av2pb.ValidatePeerCertificateChainReq_CONNECT_TO_GOOGLE)
+	creds, err := NewClientCreds(clientS2AAddress, localIdentity, s2av2pb.ValidatePeerCertificateChainReq_CONNECT_TO_GOOGLE, nil)
```
matthewstevenson88, line 133: Are we going to test the fallback in the e2e tests?

s2a_options.go
```diff
@@ -121,6 +125,34 @@ type ClientOptions struct {
 	// VerificationMode specifies the mode that S2A must use to verify the
 	// peer certificate chain.
 	VerificationMode VerificationModeType
+
+	// S2Av2 only.
```
matthewstevenson88, line 129: nit: Let's avoid the S2Av2 distinction here please. :)

Similarly throughout.

s2a_options.go
```diff
@@ -121,6 +125,34 @@ type ClientOptions struct {
...
+	// Optional fallback after dialing with S2Av2 fails.
+	FallbackOpts *FallbackOptions
+}
+
+// FallbackOptions contains the fallback options for S2Av2.
+type FallbackOptions struct {
+	// FallbackClientHandshakeFunc is used to specify fallback behavior when calling s2a.NewClientCreds().
+	// It will be called by s2av2TransportCreds's ClientHandshake function, after handshake with S2Av2 fails.
```
matthewstevenson88, line 137: s2av2TransportCreds is an internal implementation detail and should not be mentioned in a public-facing interface.

s2a_options.go
```diff
@@ -121,6 +125,34 @@ type ClientOptions struct {
 	// VerificationMode specifies the mode that S2A must use to verify the
 	// peer certificate chain.
 	VerificationMode VerificationModeType
+
+	// S2Av2 only.
+	// Optional fallback after dialing with S2Av2 fails.
+	FallbackOpts *FallbackOptions
+}
+
+// FallbackOptions contains the fallback options for S2Av2.
```
matthewstevenson88, line 134: I think we need to say a bit more here, e.g. "FallbackOptions prescribes the fallback logic that should be taken if the application fails to connect to S2A."

s2a_options.go
```diff
@@ -121,6 +125,34 @@ type ClientOptions struct {
...
+type FallbackOptions struct {
```
matthewstevenson88, line 135: Please specify what happens if both fields are set.

s2a_options.go
```diff
@@ -121,6 +125,34 @@ type ClientOptions struct {
...
+	// FallbackClientHandshakeFunc is used to specify fallback behavior when calling s2a.NewClientCreds().
+	// It will be called by s2av2TransportCreds's ClientHandshake function, after handshake with S2Av2 fails.
+	//     originServer: the original server attempted with S2Av2.
```
matthewstevenson88, line 138: I'm afraid this terminology is confusing: could we instead say "serverAddress" or "targetAddress"? Since this is the address of the server we're trying to connect to?

Similarly, could "originConn" be renamed to "conn" or "serverConn".

Using the term "origin" or "original" implies that there is necessarily a difference, when that's not the case (it just happens to be the case for the default implementation).

s2a_options.go
```diff
@@ -121,6 +125,34 @@ type ClientOptions struct {
...
+	//     originConn: the original raw tcp connection passed into S2Av2's ClientHandshake func.
+	//                 If fallback is successful, the `originConn` should be closed.
+	//     originErr: the error encountered when performing client handshake with S2Av2.
+	// This fallback func should return a post-handshake TLS connection, plus its auth info.
```
matthewstevenson88, line 142: nit: Please remove "post-handshake". From the user perspective, this should just be a TLS connection and they should not need to reason about the protocol details.

fallback/s2a_fallback.go
```diff
@@ -0,0 +1,159 @@
...
+	"crypto/tls"
+	"fmt"
+	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/grpclog"
+	"net"
+)
+
+const (
+	alpnProtoStrH2   = "h2"
+	alpnProtoStrHttp = "http/1.1"
```
matthewstevenson88, line 33: Not saying this is a bad thing, I just want to understand: why do we need to support HTTP/1.1? If this is only being used to talk to Google, then HTTP/2 should be sufficient, no?

fallback/s2a_fallback.go
```diff
@@ -0,0 +1,159 @@
...
+	defaultHttpsPort = "443"
+)
+
+// FallbackClientHandshake establishes a TLS connection and returns it, plus its auth info.
+// Input:
```
matthewstevenson88, line 38: micro-nit: s/Input/Inputs

fallback/s2a_fallback.go
```diff
@@ -0,0 +1,159 @@
...
+//
+//	targetServer: the server attempted with S2A.
+//	conn: the raw tcp connection passed into S2A's ClientHandshake func.
```
matthewstevenson88, line 41: nit: For clarity, it may be good to say "the tcp connection to the server at address targetServer that was ..."

fallback/s2a_fallback.go
```diff
@@ -0,0 +1,159 @@
...
+//	            If fallback is successful, the `conn` should be closed.
+//	err: the error encountered when performing client handshake with S2A.
```
matthewstevenson88, line 43: micro-nit: s/client/the client-side TLS

fallback/s2a_fallback.go
```diff
@@ -0,0 +1,159 @@
...
+type FallbackClientHandshake func(ctx context.Context, targetServer string, conn net.Conn, err error) (net.Conn, credentials.AuthInfo, error)
+
+// DefaultFallbackClientHandshakeFunc returns an implementation of FallbackOptions.FallbackClientHandshakeFunc,
```
matthewstevenson88, line 46: Suggested edit: "DefaultFallbackClientHandshakeFunc returns a FallbackClientHandshake function, which establishes ..."

fallback/s2a_fallback.go
```diff
@@ -0,0 +1,159 @@
...
+// which is of type FallbackClientHandshake.
+// It establishes a TLS connection with the provided fallbackAddr, returns the new connection and its auth info.
```
matthewstevenson88, line 48: micro-nit: s/with the/to the

fallback/s2a_fallback.go
```diff
@@ -0,0 +1,159 @@
...
+//
+//	transportCreds, _ = s2a.NewClientCreds(&s2a.ClientOptions{
+//		S2AAddress: s2aAddress,
+//		EnableV2:   true,
+//		FallbackOpts: &s2a.FallbackOptions{ // optional
+//			FallbackClientHandshakeFunc: fallback.DefaultFallbackClientHandshakeFunc(fallbackAddr),
+//		},
+//	})
+//
+// The fallback server's certificate should be verifiable using OS root store.
```
matthewstevenson88, line 59: micro-nit: s/should/must

fallback/s2a_fallback.go
```diff
@@ -0,0 +1,159 @@
...
+// The fallbackAddr is expected to be a network address, e.g. example.com:port. If port is not specified,
+// it uses default port 443.
+// In the returned function's TLS config, ClientSessionCache is explicitly set to nil to disable TLS resumption,
+// and min TLS version is set to 1.3.
+func DefaultFallbackClientHandshakeFunc(fallbackAddr string) (func(context.Context, string, net.Conn, error) (net.Conn, credentials.AuthInfo, error), error) {
```
matthewstevenson88, line 64: In the function declaration, can we write the return type as FallbackClientHandshake instead of repeating?

fallback/s2a_fallback.go
```diff
@@ -0,0 +1,159 @@
...
+		return nil, err
+	}
+	return func(ctx context.Context, targetServer string, conn net.Conn, s2aErr error) (net.Conn, credentials.AuthInfo, error) {
+		fallbackTLSConfig := tls.Config{
+			MinVersion:         tls.VersionTLS13,
+			ClientSessionCache: nil,
+			NextProtos:         []string{alpnProtoStrH2},
+		}
+		fallbackDialer := &tls.Dialer{Config: &fallbackTLSConfig}
+		fbConn, fbErr := fallbackDialer.DialContext(ctx, "tcp", fallbackServerAddr)
```
matthewstevenson88, line 79: nit: the usual google3 Go convention is to use the variable name `err` for all errors.

fallback/s2a_fallback.go
```diff
@@ -0,0 +1,159 @@
...
+	if err != nil {
+		if grpclog.V(1) {
+			grpclog.Infof("error processing fallback address [%s]: %v", fallbackAddr, err)
+		}
+		return nil, "", err
+	}
+	fallbackTLSConfig := tls.Config{
+		MinVersion:         tls.VersionTLS13,
+		ClientSessionCache: nil,
+		NextProtos:         []string{alpnProtoStrHttp},
```
matthewstevenson88, line 138: Is there a reason we're only supporting HTTP/1.1 here?

s2a.go
```diff
@@ -344,3 +349,63 @@ func getVerificationMode(verificationMode VerificationModeType) s2av2pb.Validate
 		return s2av2pb.ValidatePeerCertificateChainReq_UNSPECIFIED
 	}
 }
+
+// NewS2ADialTLSContextFunc returns a dial func which establishes an MTLS connection using S2A.
+// example use with http.RoundTripper:
```
matthewstevenson88, line 354: micro-nit: s/example/Example

s2a.go
```diff
@@ -344,3 +349,63 @@ func getVerificationMode(verificationMode VerificationModeType) s2av2pb.Validate
...
+//
+//		dialTLSContext := s2a.NewS2aDialTLSContextFunc(&s2a.ClientOptions{
+//			S2AAddress:         s2aAddress, // required
+//			EnableV2:           true, // must be true
+//		})
+//	 	trans := http.DefaultTransport
```
matthewstevenson88, line 360: micro-nit: s/trans/transport

s2a.go
```diff
@@ -344,3 +349,63 @@ func getVerificationMode(verificationMode VerificationModeType) s2av2pb.Validate
...
+func NewS2ADialTLSContextFunc(opts *ClientOptions) func(ctx context.Context, network, addr string) (net.Conn, error) {
+
+	return func(ctx context.Context, network, addr string) (net.Conn, error) {
+
+		fallback := func(err error) (net.Conn, error) {
+			if opts.FallbackOpts != nil && opts.FallbackOpts.FallbackDialer != nil &&
+				opts.FallbackOpts.FallbackDialer.Dialer != nil && opts.FallbackOpts.FallbackDialer.ServerAddr != "" {
+				fbDialer := opts.FallbackOpts.FallbackDialer
+				grpclog.Infof("fall back to dial: %s", fbDialer.ServerAddr)
+				fbConn, fbErr := fbDialer.Dialer.DialContext(ctx, network, fbDialer.ServerAddr)
```
matthewstevenson88, line 371: micro-nit: s/fbErr/err

s2a.go
```diff
@@ -344,3 +349,63 @@ func getVerificationMode(verificationMode VerificationModeType) s2av2pb.Validate
...
+				if fbErr != nil {
+					return nil, fmt.Errorf("error fallback to %s: %v; S2A error: %w", fbDialer.ServerAddr, fbErr, err)
+				} else {
+					return fbConn, nil
+				}
+			}
+			return nil, err
+		}
+
+		factory, e := NewTLSClientConfigFactory(opts)
```
matthewstevenson88, line 381: micro-nit: s/e/err

s2a.go
```diff
@@ -344,3 +349,63 @@ func getVerificationMode(verificationMode VerificationModeType) s2av2pb.Validate
...
+		if e != nil {
+			grpclog.Infof("error creating S2A client config factory: %v", e)
+			return fallback(e)
+		}
+
+		serverName, _, e := net.SplitHostPort(addr)
```
matthewstevenson88, line 387: micro-nit: s/e/err

s2a.go
```diff
@@ -344,3 +349,63 @@ func getVerificationMode(verificationMode VerificationModeType) s2av2pb.Validate
...
+		if e != nil {
+			serverName = addr
+		}
+		s2aTLSConfig, e := factory.Build(ctx, &TLSClientConfigOptions{
```
matthewstevenson88, line 391: micro-nit: s/e/err

s2a.go
```diff
@@ -344,3 +349,63 @@ func getVerificationMode(verificationMode VerificationModeType) s2av2pb.Validate
...
+		})
+		if e != nil {
+			grpclog.Infof("error building S2A TLS config: %v", e)
+			return fallback(e)
+		}
+
+		s2aDialer := &tls.Dialer{
+			Config: s2aTLSConfig,
+		}
+		c, e := s2aDialer.DialContext(ctx, network, addr)
```
matthewstevenson88, line 402: micro-nit: s/e/err

s2a.go
```diff
@@ -344,3 +349,63 @@ func getVerificationMode(verificationMode VerificationModeType) s2av2pb.Validate
...
+		if e != nil {
+			grpclog.Infof("error dialing with S2A to %s: %v", addr, e)
+			return fallback(e)
+		} else {
```
matthewstevenson88, line 406: Usual convention is to omit the last else when returning from the function.

matthewstevenson88 (commented):
Thanks for the PR Kui! Generally LGTM, just a large handful of very small nits. :)

s2a.go
xmenxk: in this case, there is an `err` from S2A failure, that's passed in to the func, and then `fbErr` is for the fallback dial, we need to differentiate the two.

fallback/s2a_fallback.go
xmenxk: in this case, there's an `s2aErr` from s2a failure, which is passed into the func, then `fbErr` is from dialing the fallback connection, I'd like the naming to clearly differentiate the two.

fallback/s2a_fallback.go
xmenxk: thanks Matt, I had the (likely outdated) thought that h2 is new, and 1.1 would have wider support...
anyway I tested again by only setting h2, it also works.
I'll add both here, so the client and server can pick one that's supported in both ends.

fallback/s2a_fallback.go
xmenxk: replied in another comment

internal/v2/s2av2_test.go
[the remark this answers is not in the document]
xmenxk: done thanks

fallback/s2a_fallback.go
matthewstevenson88: SGTM, makes sense.

fallback/s2a_fallback.go
matthewstevenson88: If we want to add both, that's fine, but here we're only setting HTTP/1.1 and above we're only setting HTTP/2. I think we need both strings in both places.

fallback/s2a_fallback_test.go
```diff
@@ -0,0 +1,101 @@
...
+}
+func TestDefaultFallbackDialerAndAddress(t *testing.T) {
+	for _, tc := range []struct {
+		desc              string
+		inputFallbackAddr string
+		dialerInitialized bool
+		expectedAddr      string
+		expectNilErr      bool
+	}{
+		{
```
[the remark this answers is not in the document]
matthewstevenson88: Friendly ping :)

fallback/s2a_fallback_test.go
```diff
@@ -0,0 +1,101 @@
...
+		},
+		{
+			"test empty fallback address",
+			"",
+			false,
+			false,
+		},
+	} {
+		fbFunc, err := DefaultFallbackClientHandshakeFunc(tc.inputFallbackAddr)
+		if got, want := fbFunc != nil, tc.funcInitialized; got != want {
```
matthewstevenson88, line 52: Sorry that I missed this before: in the success cases, I'm concerned that we're not adequately testing fbFunc, since we're just checking that it is non-nil. In the success cases, can we also check that fbFunc behaves as expected e.g. with fake/mock inputs?

fallback/s2a_fallback_test.go
```diff
@@ -0,0 +1,101 @@
...
+		{
+			"test empty fallback address",
+			"",
+			false,
+			"",
+			false,
+		},
+	} {
+		fbDialer, fbAddr, err := DefaultFallbackDialerAndAddress(tc.inputFallbackAddr)
+		if got, want := fbDialer != nil, tc.dialerInitialized; got != want {
```
matthewstevenson88, line 91: Same here RE testing the success cases with fake/mock inputs.

fallback/s2a_fallback_test.go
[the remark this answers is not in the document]
xmenxk: thanks Matt, so in the code comments, we made it clear that the function expects a valid network address as input, and we don't do url parsing to remove schemes such as `http` `https`. I hope we don't have to get into the  url parsing/sanitization logic.
The user is responsible to pass in a valid address. 
I also added a few more test cases, where the addresses can be ip/ipv6  as well. let me know what you think thanks:)

fallback/s2a_fallback.go
xmenxk: I just updated the code, 
for the fallback handshake function: I'm setting only `h2`, reason is that this is for GRPC and it speaks http2 only

for the fallback dialer function: which is for the http client use case. I'm setting both `h2` and `http 1.1`, where client and server can negotiate into one protocol they both like 

plz correct me if I misunderstood, thanks again!

fallback/s2a_fallback_test.go
matthewstevenson88: I think there are ways we can test *parts* of this without relying on a starting up a fake server, but for a complete success test I agree we need e2e tests. For example,

- Factor out the creation of TLS config (which is duplicated code) and unit test this function independently.
- For `DefaultFallbackDialerAndAddress`, we can check that the returned tls.Dialer is exactly the struct we expect.
- Create a non-exported function defaultFallbackClientHandshakeFuncInternal that allows us to inject the functionality in lines 77-78, for example:

```
func DefaultFallbackClientHandshakeFunc(fallbackAddr string) (FallbackClientHandshake, error) ... {
 // call defaultFallbackClientHandshakeFuncInternal and pass in internalDialer
}

func defaultFallbackClientHandshakeFuncInternal(fallbackAddr string, config *tls.Config, dialer func(ctx, string, *tls.config) (net.Conn, error)) (FallbackClientHandshake, error) ... {
  // current implementation but call dialer instead of lines 77-78
}

func internalDialer(ctx context.Context, fallbackAddr string, config *tls.Config) (net.Conn, error) {
  fallbackDialer := &tls.Dialer{Config: config}
  return fallbackDialer.DialContext(ctx, "tcp", fallbackServerAddr)
}
```

WDYT?

fallback/s2a_fallback.go
```diff
@@ -0,0 +1,169 @@
...
+// It supports GRPC use case, thus the alpn is set to 'h2'.
+var FallbackTLSConfigGRPC = tls.Config{
+	MinVersion:         tls.VersionTLS13,
+	ClientSessionCache: nil,
+	NextProtos:         []string{alpnProtoStrH2},
+}
+
+// FallbackTLSConfigHTTP is a tls.Config used by the DefaultFallbackDialerAndAddress func.
+// It supports the HTTP use case and the alpn is set to both 'http/1.1' and 'h2'.
+var FallbackTLSConfigHTTP = tls.Config{
```
matthewstevenson88, line 48: nit: Does this need to be exported?

fallback/s2a_fallback.go
```diff
@@ -0,0 +1,169 @@
...
+
+const (
+	alpnProtoStrH2   = "h2"
+	alpnProtoStrHttp = "http/1.1"
+	defaultHttpsPort = "443"
+)
+
+// FallbackTLSConfigGRPC is a tls.Config used by the DefaultFallbackClientHandshakeFunc function.
+// It supports GRPC use case, thus the alpn is set to 'h2'.
+var FallbackTLSConfigGRPC = tls.Config{
```
matthewstevenson88, line 40: nit: Does this need to be exported?

matthewstevenson88 (commented):
LGTM modulo the failing hygiene test. Just give me a ping when it's fixed and I'll approve. Thanks again for the PR Kui! :)

fallback/s2a_fallback.go
xmenxk: I need to be able to override this in e2e tests, so setting it up here

fallback/s2a_fallback.go
xmenxk: same as above, thx

matthewstevenson88 (approved):

PR #92, opened 2023-03-07
e2e tests for fallback logic

xmenxk:
for GRPC, add e2e test for fallback logic
for HTTP, add e2e tests for both working and fallback

### DIFF
--- a/internal/v2/s2av2_e2e_test.go
+++ b/internal/v2/s2av2_e2e_test.go
@@ -132,7 +132,7 @@ func startServer(t *testing.T, s2aAddress string, localIdentities []*commonpbv1.
 	return lis.Addr().String()
 }
 
-// startFallbackServer runs a GRPC echo testing server.
+// startFallbackServer runs a GRPC echo testing server and returns the address.
 // It's used to test the default fallback logic upon S2A failure.
 func startFallbackServer(t *testing.T) string {
 	lis, err := net.Listen("tcp", ":0")
@@ -143,7 +143,10 @@ func startFallbackServer(t *testing.T) string {
 	if err != nil {
 		t.Errorf("failure initializing tls.certificate: %v", err)
 	}
+	// Client certs are not required for the fallback server.
 	creds := credentials.NewTLS(&tls.Config{
+		MinVersion:   tls.VersionTLS13,
+		MaxVersion:   tls.VersionTLS13,
 		Certificates: []tls.Certificate{cert},
 	})
 	s := grpc.NewServer(grpc.Creds(creds))
@@ -288,13 +291,8 @@ func TestEndToEndUsingFakeS2AOnUDSEmptyId(t *testing.T) {
 }
 
 func TestGRPCFallbackEndToEndUsingFakeS2AOverTCP(t *testing.T) {
-	fallback.FallbackTLSConfigGRPC = tls.Config{
-		MinVersion:         tls.VersionTLS13,
-		ClientSessionCache: nil,
-		NextProtos:         []string{"h2"},
-		// set for testing only
-		InsecureSkipVerify: true,
-	}
+	// Set for testing only.
+	fallback.FallbackTLSConfigGRPC.InsecureSkipVerify = true
 	os.Setenv(accessTokenEnvVariable, "TestE2ETCP_token")
 	// Start the fake S2A for the server.
 	serverS2AAddr := startFakeS2A(t, "TestE2ETCP_token")
@@ -320,12 +318,21 @@ func TestGRPCFallbackEndToEndUsingFakeS2AOverTCP(t *testing.T) {
 	if err != nil {
 		t.Errorf("error creating fallback handshake function: %v", err)
 	}
+	fallbackCalled := false
+	fallbackHandshakeWrapper := func(ctx context.Context, targetServer string, conn net.Conn, err error) (net.Conn, credentials.AuthInfo, error) {
+		fallbackCalled = true
+		return fallbackHandshake(ctx, targetServer, conn, err)
+	}
 	// Set wrong S2A address for client to trigger S2A failure and fallback.
 	runClient(ctx, t, "not_exist", serverAddr, &commonpbv1.Identity{
 		IdentityOneof: &commonpbv1.Identity_Hostname{
 			Hostname: "test_rsa_client_identity",
 		},
-	}, fallbackHandshake)
+	}, fallbackHandshakeWrapper)
+
+	if !fallbackCalled {
+		t.Errorf("fallbackHandshake is not called")
+	}
 }
 
 func TestNewClientTlsConfigWithTokenManager(t *testing.T) {
--- a/internal/v2/s2av2_e2e_test.go
+++ b/internal/v2/s2av2_e2e_test.go
@@ -23,7 +23,6 @@ import (
 	"context"
 	"crypto/tls"
 	"fmt"
-	"github.com/google/s2a-go/internal/tokenmanager"
 	"io/ioutil"
 	"net"
 	"os"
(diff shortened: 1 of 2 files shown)


internal/v2/s2av2_e2e_test.go
```diff
@@ -128,9 +132,33 @@ func startServer(t *testing.T, s2aAddress string, localIdentities []*commonpbv1.
 	return lis.Addr().String()
 }
 
+// startFallbackServer runs a GRPC echo testing server.
```
matthewstevenson88, line 135: nit: Please mention that it returns the address on which the server is listening.

internal/v2/s2av2_e2e_test.go
```diff
@@ -128,9 +132,33 @@ func startServer(t *testing.T, s2aAddress string, localIdentities []*commonpbv1.
...
+	lis, err := net.Listen("tcp", ":0")
+	if err != nil {
+		t.Errorf("net.Listen(tcp, :0) failed: %v", err)
+	}
+	cert, err := tls.X509KeyPair(serverCertpem, serverKeypem)
+	if err != nil {
+		t.Errorf("failure initializing tls.certificate: %v", err)
+	}
+	creds := credentials.NewTLS(&tls.Config{
+		Certificates: []tls.Certificate{cert},
```
matthewstevenson88, line 147: Should we hardcode the min/max TLS version in this tls.Config to 1.3 to make sure that the client is negotiating TLS 1.3?

internal/v2/s2av2_e2e_test.go
```diff
@@ -128,9 +132,33 @@ func startServer(t *testing.T, s2aAddress string, localIdentities []*commonpbv1.
...
+func startFallbackServer(t *testing.T) string {
+	lis, err := net.Listen("tcp", ":0")
+	if err != nil {
+		t.Errorf("net.Listen(tcp, :0) failed: %v", err)
+	}
+	cert, err := tls.X509KeyPair(serverCertpem, serverKeypem)
+	if err != nil {
+		t.Errorf("failure initializing tls.certificate: %v", err)
+	}
+	creds := credentials.NewTLS(&tls.Config{
```
matthewstevenson88, line 146: Please add a comment that explains that we intentionally are not requesting client certs. :)

internal/v2/s2av2_e2e_test.go
```diff
@@ -256,7 +284,48 @@ func TestEndToEndUsingFakeS2AOnUDSEmptyId(t *testing.T) {
 	// Finally, start up the client.
 	ctx, cancel := context.WithTimeout(context.Background(), defaultE2ETimeout)
 	defer cancel()
-	runClient(ctx, t, clientS2AAddr, serverAddr, nil)
+	runClient(ctx, t, clientS2AAddr, serverAddr, nil, nil)
+}
+
+func TestGRPCFallbackEndToEndUsingFakeS2AOverTCP(t *testing.T) {
+	fallback.FallbackTLSConfigGRPC = tls.Config{
```
matthewstevenson88, line 291: Rather than totally overwriting fallback.FallbackTLSConfigGRPC, can we just set the value of the `InsecureSkipVerify` field?

internal/v2/s2av2_e2e_test.go
```diff
@@ -256,7 +284,48 @@ func TestEndToEndUsingFakeS2AOnUDSEmptyId(t *testing.T) {
...
+
+	// Finally, start up the client.
+	ctx, cancel := context.WithTimeout(context.Background(), defaultE2ETimeout)
+	defer cancel()
+	fallbackHandshake, err := fallback.DefaultFallbackClientHandshakeFunc(fallbackServerAddr)
+	if err != nil {
+		t.Errorf("error creating fallback handshake function: %v", err)
+	}
+	// Set wrong S2A address for client to trigger S2A failure and fallback.
+	runClient(ctx, t, "not_exist", serverAddr, &commonpbv1.Identity{
```
matthewstevenson88, line 324: In this test, how do we know the fallback was invoked?

s2a_e2e_test.go
```diff
@@ -405,3 +410,101 @@ func TestNewTLSClientConfigFactoryWithoutTokenManager(t *testing.T) {
...
+	}
+	respBody, err := io.ReadAll(resp.Body)
+	if err != nil {
+		t.Errorf("error reading HTTP response: %v", err)
+	}
+	if got, want := string(respBody), "hello"; got != want {
+		t.Errorf("expecting HTTP response:[%s], got [%s]", want, got)
+	}
+}
+func TestHTTPEndToEndUsingFakeS2AOverTCP(t *testing.T) {
```
matthewstevenson88, line 465: Can we also test the gRPC APIs at this layer?

s2a_e2e_test.go
```diff
@@ -405,3 +410,101 @@ func TestNewTLSClientConfigFactoryWithoutTokenManager(t *testing.T) {
 		t.Errorf("tls.Config has unexpected certificate: got: %v, want: %v", got, want)
 	}
 }
+
+// startHTTPServer runs an HTTP server on a random local port and serves a /hello endpoint.
```
matthewstevenson88, line 414: Is there a reason why we only have HTTP tests at the top layer and only have GRPC tests in the v2 layer, rather than having both kinds of tests at both layers?

s2a_e2e_test.go
```diff
@@ -405,3 +410,101 @@ func TestNewTLSClientConfigFactoryWithoutTokenManager(t *testing.T) {
...
+	// Start the server.
+	serverAddr := startHTTPServer(t)
+	t.Logf("HTTP server running at address: %v", serverAddr)
+
+	// Finally, start up the client.
+	runHTTPClient(t, clientHandshakerAddr, serverAddr, nil)
+}
+
+func TestHTTPFallbackEndToEndUsingFakeS2AOverTCP(t *testing.T) {
+	fallback.FallbackTLSConfigHTTP = tls.Config{
```
matthewstevenson88, line 481: Same here RE overwriting the config vs. just setting 1 field.

s2a_e2e_test.go
```diff
@@ -405,3 +410,101 @@ func TestNewTLSClientConfigFactoryWithoutTokenManager(t *testing.T) {
...
+	// Start the fake S2As for the client.
+	clientHandshakerAddr := startFakeS2A(t, true, testV2AccessToken)
+	t.Logf("fake handshaker for client running at address: %v", clientHandshakerAddr)
+
+	// Start the server.
+	serverAddr := startHTTPServer(t)
+	t.Logf("HTTP server running at address: %v", serverAddr)
+
+	// Finally, start up the client.
+	runHTTPClient(t, clientHandshakerAddr, serverAddr, nil)
```
matthewstevenson88, line 477: Same here and below RE how do we know the fallback was invoked successfully.

matthewstevenson88 (commented):
Thanks for the PR Kui!

s2a_e2e_test.go
xmenxk: done!

internal/v2/s2av2_e2e_test.go
xmenxk: added thanks

internal/v2/s2av2_e2e_test.go
xmenxk: I added a wrapper to the fallback function, which flips a bool var to indicate being called.

s2a_e2e_test.go
xmenxk: I made the target and fallback  http servers return different message.
By verifying the response returned is from the fallback server, we know fallback is called.

s2a_e2e_test.go
xmenxk: I added the tests to both layers fro grpc. For http case, it doesn't call into internal code, that's why the test is only at the top level.

s2a_e2e_test.go
```diff
@@ -405,3 +472,105 @@ func TestNewTLSClientConfigFactoryWithoutTokenManager(t *testing.T) {
 		t.Errorf("tls.Config has unexpected certificate: got: %v, want: %v", got, want)
 	}
 }
+
+// startHTTPServer runs an HTTP server on a random local port and serves a /hello endpoint.
+// The response of the /hello endpoint should be passed in via the `resp` parameter.
+// It returns the address of the server.
+func startHTTPServer(t *testing.T, resp string) string {
+	cert, _ := tls.X509KeyPair(serverCertpem, serverKeypem)
+	tlsConfig := tls.Config{
```
matthewstevenson88, line 481: Can we set min/max TLS versions to 1.3 here?

matthewstevenson88 (approved):
LGTM with 1 minor comment. :)

PR #93, opened 2023-03-08
add EnableLegacyMode to s2a_options

xmenxk:
this is step one to make v2 the default and deprecate the `EnableV2` flag

### DIFF
--- a/s2a_options.go
+++ b/s2a_options.go
@@ -120,6 +120,8 @@ type ClientOptions struct {
 	//
 	// If true, enables the use of S2Av2.
 	EnableV2 bool
+	// If true, enable the use of legacy S2Av1.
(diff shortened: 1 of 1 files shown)


s2a_options.go
```diff
@@ -120,6 +120,8 @@ type ClientOptions struct {
 	//
 	// If true, enables the use of S2Av2.
 	EnableV2 bool
+	// If true, enables the use of legacy S2Av1.
+	EnableLegacyMode bool
```
matthewstevenson88, line 124: Should we add logic to explicitly disable v2 if EnableLegacyMode is set to true?

matthewstevenson88: SGTM.

PR #96, opened 2023-03-16
remove EnableV2 flag

xmenxk:
v2 is now the default

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -32,7 +32,7 @@ go_library(
         "s2a_options.go",
         "s2a_utils.go",
     ],
-    importpath = "github.com/google/s2a-go/s2a",
+    importpath = "github.com/google/s2a-go",
     deps = [
         "//fallback:s2a_fallback",
         "//internal/handshaker",
--- a/BUILD
+++ b/BUILD
@@ -32,7 +32,7 @@ go_library(
         "s2a_options.go",
         "s2a_utils.go",
     ],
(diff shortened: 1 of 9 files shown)


BUILD
```diff
@@ -32,7 +32,7 @@ go_library(
         "s2a_options.go",
         "s2a_utils.go",
     ],
-    importpath = "github.com/google/s2a-go/s2a",
+    importpath = "github.com/google/s2a-go",
```
matthewstevenson88, line 35: Why does this need to change in this PR?

xmenxk: I updated the TODO by rmehta19 in the example directory to call s2a.NewCredentials instead of the internal v2 method, and found that the import path needs correction.

matthewstevenson88: It's not clear to me why this importpath would need updating - can you clarify?

The reason I'm hesitant to change this line is because this is a breaking change for any current users of this repo.

xmenxk: I see, in that case, I restored the change

xmenxk: wait, 
if we keep the import path as: "github.com/google/s2a-go/s2a"
now the issue is kokoro tests are failing because:

example/client/client.go:25:2: no required module provides package github.com/google/s2a-go/s2a; to add it: go get github.com/google/s2a-go/s2a

matthewstevenson88 (approved):

PR #98, opened 2023-03-16
shortor timeout when connecting to s2a service during handshake

xmenxk:
also added a flag to make it configurable

### DIFF
--- a/internal/v2/s2av2.go
+++ b/internal/v2/s2av2.go
@@ -45,7 +45,7 @@ const (
 	s2aSecurityProtocol = "tls"
 )
 
-var connectS2ATimeout = flag.Duration("connect_s2a_timeout", 3*time.Second, "Timeout enforced when connecting to the S2A service.")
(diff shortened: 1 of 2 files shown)


internal/v2/s2av2.go
```diff
@@ -119,7 +121,7 @@ func (c *s2av2TransportCreds) ClientHandshake(ctx context.Context, serverAuthori
 	}
 	// Remove the port from serverAuthority.
 	serverName := removeServerNamePort(serverAuthority)
-	ctx, cancel := context.WithTimeout(ctx, defaultTimeout)
+	ctx, cancel := context.WithTimeout(ctx, *connectS2ATimeout)
```
matthewstevenson88, line 124: Do we need the timeout in the HTTP APIs as well?

xmenxk: good catch, added thanks

matthewstevenson88 (commented):
Sending back since it seems there are a few test failures.

matthewstevenson88 (approved):

Issue #105, opened 2023-04-27
BUILD.bazel being ignored causes problems with bazel projects with vendored libraries


FranciscoKurpiel:
On [.gitignore](https://github.com/google/s2a-go/blob/main/.gitignore) there is an instruction to ignore `BUILD.bazel`. This is a problem for our project because we vendor our libraries, and we use bazel + gazelle, because the files generated by gazelle will not be tracked by git. When the code is checked-out from git it will be missing the files and and bazel will fail.

### How to reproduce:
- create a new go project
- add `github.com/google/s2a-go` as a dependency
- `go mod tidy && go mod vendor`
- configure bazel with gazelle
- `bazel run //:gazelle`
- `bazel test //...` # will run and show that there are not tests to run
- `git commit -am "initial"` # CAUSE OF THE ISSUE HERE: all BUILD.bazel files generated by bazel on 
`vendor/github.com/google/s2a-go` will not be tracked
- `git push ...`

On another machine:
- `git clone ...`
- `bazel test //...` # << unexpected behavior here

### Expected result:
The same result as on the machine that created the git repository: bazel should run and say that there are no tests to run

### Observed result:
An error like the one here:
```
Starting local Bazel server and connecting to it...
INFO: Invocation ID: 5e938a9f-6e8b-4dfa-ad90-37fa2720ac22
ERROR: /home/xyko/tmp/events3/vendor/google.golang.org/api/internal/BUILD.bazel:3:11: no such package 'vendor/github.com/google/s2a-go/fallback': BUILD file not found in any of the following directories. Add a BUILD file to a directory to mark it as a package.
 - /home/xyko/tmp/events3/vendor/github.com/google/s2a-go/fallback and referenced by '//vendor/google.golang.org/api/internal:internal'
ERROR: Analysis of target '//vendor/google.golang.org/api/internal:internal' failed; build aborted: Analysis failed
INFO: Elapsed time: 3.812s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (522 packages loaded, 329 targets configured)
ERROR: Couldn't start the build. Unable to run tests
```

### Comments
- I believe the library should NOT list `BUILD.bazel` on `.gitignore`.
- The issue affect indirect users of this library, like users of `google.golang.org/api`

ramilmsh:
same here. I updated some google package and it included this as a dependency and now my builds are failing

xmenxk:
Thanks, I updated the gitignore file in v0.1.3

[...]

PR #106, opened 2023-04-28
fix vulnerable dependencies

xmenxk:
- upgrade a couple dependency modules which github finds vulnerable.
- also changed .gitignore to not ignore bazel files, since some customers want to check those into repo https://github.com/google/s2a-go/issues/105

go build ./...
go test ./...


matthewstevenson88 (approved):
Thank you for the fix!

PR #108, opened 2023-05-08
Add option to inject custom bidirectional stream to S2A.

zeromath:
This is needed to allow user to configure the server authorization policy and the stream to talk to the S2A server.

### DIFF
--- a/s2a_options.go
+++ b/s2a_options.go
@@ -26,6 +26,7 @@ import (
 	"github.com/google/s2a-go/fallback"
 
 	s2apb "github.com/google/s2a-go/internal/proto/common_go_proto"
+	s2av2pb "github.com/google/s2a-go/internal/proto/v2/s2a_go_proto"
 )
 
 // Identity is the interface for S2A identities.
@@ -78,6 +79,14 @@ const (
 	Spiffe
 )
 
+// s2AStream defines the operation for communicating with the S2A server over a bidirectional stream.
+type s2AStream interface {
+	// Send sends the message to the S2A server.
+	Send(*s2av2pb.SessionReq) error
+	// Recv receives the message from the S2A server.
+	Recv() (*s2av2pb.SessionResp, error)
+}
+
 // ClientOptions contains the client-side options used to establish a secure
 // channel using the S2A handshaker service.
 type ClientOptions struct {
@@ -125,6 +134,12 @@ type ClientOptions struct {
 
 	// Optional fallback after dialing with S2A fails.
 	FallbackOpts *FallbackOptions
+
+	// Generates an s2AStream interface for talking to the S2A server.
+	getS2AStream func() (*s2AStream, error)
+
+	// Serialized user specified policy for server authorization.
+	serverAuthorizationPolicy []byte
 }
 
 // FallbackOptions prescribes the fallback logic that should be taken if the application fails to connect with S2A.
@@ -170,6 +185,9 @@ type ServerOptions struct {
 	// VerificationMode specifies the mode that S2A must use to verify the
 	// peer certificate chain.
 	VerificationMode VerificationModeType
+
+	// Generates an s2AStream interface for talking to the S2A server.
+	getS2AStream func() (*s2AStream, error)
 }
 
 // DefaultServerOptions returns the default server options.


s2a_options.go
```diff
@@ -125,6 +145,12 @@ type ClientOptions struct {
 
 	// Optional fallback after dialing with S2A fails.
 	FallbackOpts *FallbackOptions
+
+	// Generates S2AStream for talking to the S2A server.
+	GetS2AStream *func() (*S2AStream, error)
```
matthewstevenson88, line 150: Can this also be unexported?

s2a_options.go
```diff
@@ -125,6 +145,12 @@ type ClientOptions struct {
 
 	// Optional fallback after dialing with S2A fails.
 	FallbackOpts *FallbackOptions
+
+	// Generates S2AStream for talking to the S2A server.
```
matthewstevenson88, line 149: micro-nit: s/S2AStream/an S2AStream instance

s2a_options.go
matthewstevenson88, line 150: Any reason we need to return `*func()...` instead of `func()...`? IIUC, the latter is more common.

s2a_options.go
```diff
@@ -170,6 +196,9 @@ type ServerOptions struct {
 	// VerificationMode specifies the mode that S2A must use to verify the
 	// peer certificate chain.
 	VerificationMode VerificationModeType
+
+	// Generates S2AStream for talking to the S2A server.
```
matthewstevenson88, line 200: Same comments here as above.

s2a_options.go
```diff
@@ -78,6 +80,24 @@ const (
 	Spiffe
 )
 
+// S2AStream defines the stream that S2A must use to connect to the S2A server.
+type S2AStream interface {
+	Send(*s2av2pb.SessionReq) error
+	Recv() (*s2av2pb.SessionResp, error)
+}
+
+type S2AGrpcStream struct {
```
matthewstevenson88, line 89: Rather than offering a gRPC implementation, I'd prefer if this is the default when `GetS2AStream` is not populated.

s2a_options.go
```diff
@@ -78,6 +80,24 @@ const (
 	Spiffe
 )
 
+// S2AStream defines the stream that S2A must use to connect to the S2A server.
+type S2AStream interface {
+	Send(*s2av2pb.SessionReq) error
```
matthewstevenson88, line 85: Please add comments for these 2 APIs.

s2a_options.go
```diff
@@ -78,6 +80,24 @@ const (
 	Spiffe
 )
 
+// S2AStream defines the stream that S2A must use to connect to the S2A server.
+type S2AStream interface {
```
matthewstevenson88, line 84: Can this also be unexported?

s2a_options.go
```diff
@@ -78,6 +80,24 @@ const (
 	Spiffe
 )
 
+// S2AStream defines the stream that S2A must use to connect to the S2A server.
```
matthewstevenson88, line 83: Suggested edit: S2AStream defines the operations for communicating with the S2A server over a bidirectional stream.

matthewstevenson88 (commented):
Thanks for the PR Zhan!

matthewstevenson88:
@xmenxk It looks like the Kokoro tests didn't run on this PR - do you happen to know why that happened?

Maybe we just got unlucky because the commits were pushed before @zeromath had the permissions to run the Kokoro tests. @zeromath Please ensure that the Kokoro tests start running after you push your next commit.

s2a_options.go
```diff
@@ -125,6 +145,12 @@ type ClientOptions struct {
 
 	// Optional fallback after dialing with S2A fails.
 	FallbackOpts *FallbackOptions
+
+	// Generates S2AStream for talking to the S2A server.
+	GetS2AStream *func() (*S2AStream, error)
+
+	// Serialized user specified policy for server authorization.
+	serverAuthorizationPolicy []byte
```
xmenxk, line 153: this probably should be exported as well

s2a_options.go
zeromath: Done.

s2a_options.go
zeromath: Per offline discussion, make all newly added fields unexported.

s2a_options.go
zeromath: Changed to `func()…`.

s2a_options.go
zeromath: Done.

s2a_options.go
zeromath: Done.

s2a_options.go
zeromath: Done.

s2a_options.go
zeromath: Done.

s2a_options.go
zeromath: Per offline discussion, all newly added fields will be unexported.

matthewstevenson88 (approved):

xmenxk (approved):

PR #112, opened 2023-05-15
Switch to use use custom Stream create function if available.

zeromath:


internal/v2/s2av2.go
```diff
@@ -194,12 +200,19 @@ func (c *s2av2TransportCreds) ServerHandshake(rawConn net.Conn) (net.Conn, crede
...
 	defer cancel()
-	cstream, err := createStream(ctx, c.s2av2Address)
+
+	var s2AStream stream.S2AStream
+	var err error
+	if c.getS2AStream == nil {
+		s2AStream, err = createStream(ctx, c.s2av2Address)
+	} else {
+		s2AStream, err = c.getS2AStream(ctx, c.s2av2Address)
+	}
```
xmenxk, line 210: nit: maybe factor out these lines of code, since it's the same between client and server handshake

xmenxk (approved):

internal/v2/s2av2.go
zeromath: Done.

I integrated this part of logic into `createStream` function.

internal/v2/s2av2.go
```diff
@@ -307,12 +307,37 @@ func removeServerNamePort(serverName string) string {
...
+
+func (x *s2AGrpcStream) Recv() (*s2av2pb.SessionResp, error) {
+	return x.stream.Recv()
+}
+
+func (x *s2AGrpcStream) CloseSend() error {
+	return x.stream.CloseSend()
+}
+
+func createStream(getS2AStream func(ctx context.Context, s2av2Address string) (stream.S2AStream, error), ctx context.Context, s2av2Address string) (stream.S2AStream, error) {
```
matthewstevenson88, line 326: Can we add unit tests that catch the changes here?

matthewstevenson88 (commented):
Thanks for the PR Zhan!

internal/v2/s2av2.go
zeromath: Isn't this an inner function and should be left untested?

I searched in `s2av2_test.go` and `s2av2_e2e_test.go` and didn't find any direct test on this function.

matthewstevenson88: Previously, this function was tested indirectly because we're creating real streams to the fake S2Av2 server.

However, now, we're not testing the changes to this function at all. So I think it is reasonable to add unit tests for this function. This should be fine to as long as the test is in the same package as the `s2av2.go` file.

zeromath: Done.

matthewstevenson88 (approved):

PR #115, opened 2023-07-26
retry if err when talking to s2a gRPC server

xmenxk:
add retry logic similar to the code used for talking to metadata server:
https://github.com/googleapis/google-cloud-go/blob/main/compute/metadata/retry.go

summary: err will be retried for up to 5 times: 
1st backoff:   random duration up to 0.1s
2nd backoff: random duration up to 0.2s
3rd backoff:  random duration up to 0.4s
4th backoff:  random duration up to 0.8s
5th backoff:  random duration up to 1.6s

tested running a test gae app and triggering the retry logic by:
case1: giving it a wrong s2a address
case2: setting a wrong server name in tls.config, so the handshake fails
and was able to observe (via log) 5 retries made for both cases.

### DIFF
--- a/internal/v2/s2av2.go
+++ b/internal/v2/s2av2.go
@@ -132,22 +132,13 @@ func (c *s2av2TransportCreds) ClientHandshake(ctx context.Context, serverAuthori
 	serverName := removeServerNamePort(serverAuthority)
 	timeoutCtx, cancel := context.WithTimeout(ctx, GetS2ATimeout())
 	defer cancel()
-	var err error
 	var s2AStream stream.S2AStream
-	retryer := NewRetryer()
-	for {
-		if err = ctx.Err(); err != nil {
-			break
-		}
-		s2AStream, err = createStream(timeoutCtx, c.s2av2Address, c.getS2AStream)
-		if backoff, shouldRetry := retryer.Retry(err); shouldRetry {
-			if sleepErr := retry.Sleep(ctx, backoff); sleepErr != nil {
-				break
-			}
-			continue
-		}
-		break
-	}
+	var err error
+	retry.Run(timeoutCtx,
+		func() error {
+			s2AStream, err = createStream(timeoutCtx, c.s2av2Address, c.getS2AStream)
+			return err
+		})
 	if err != nil {
 		grpclog.Infof("Failed to connect to S2Av2: %v", err)
 		if c.fallbackClientHandshake != nil {
@@ -172,20 +163,11 @@ func (c *s2av2TransportCreds) ClientHandshake(ctx context.Context, serverAuthori
 	if c.serverName != "" {
 		sn = c.serverName
 	}
-	retryer = NewRetryer()
-	for {
-		if err = ctx.Err(); err != nil {
-			break
-		}
-		config, err = tlsconfigstore.GetTLSConfigurationForClient(sn, s2AStream, tokenManager, c.localIdentity, c.verificationMode, c.serverAuthorizationPolicy)
-		if backoff, shouldRetry := retryer.Retry(err); shouldRetry {
-			if sleepErr := retry.Sleep(ctx, backoff); sleepErr != nil {
-				break
-			}
-			continue
-		}
-		break
-	}
+	retry.Run(timeoutCtx,
+		func() error {
+			config, err = tlsconfigstore.GetTLSConfigurationForClient(sn, s2AStream, tokenManager, c.localIdentity, c.verificationMode, c.serverAuthorizationPolicy)
+			return err
+		})
 	if err != nil {
 		grpclog.Info("Failed to get client TLS config from S2Av2: %v", err)
 		if c.fallbackClientHandshake != nil {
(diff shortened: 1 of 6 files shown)


s2a.go
```diff
@@ -390,9 +391,21 @@ func NewS2ADialTLSContextFunc(opts *ClientOptions) func(ctx context.Context, net
...
 		timeoutCtx, cancel := context.WithTimeout(ctx, v2.GetS2ATimeout())
 		defer cancel()
-		s2aTLSConfig, err := factory.Build(timeoutCtx, &TLSClientConfigOptions{
-			ServerName: serverName,
-		})
+
+		var s2aTLSConfig *tls.Config
+		retryer := retry.NewRetryer()
+		for {
+			s2aTLSConfig, err = factory.Build(timeoutCtx, &TLSClientConfigOptions{
```
matthewstevenson88, line 398: Shouldn't we be checking the context to see if we've hit the timeout before re-entering? Similarly throughout.

matthewstevenson88 (commented):
In general I'm OK with the retry logic as we've structured it here, but please add tests for the changes in s2av2.go and s2a.go that exercise this logic in both success and failure cases.

s2a.go
xmenxk: check added, thanks

internal/v2/s2av2.go
```diff
@@ -131,7 +132,22 @@ func (c *s2av2TransportCreds) ClientHandshake(ctx context.Context, serverAuthori
 	serverName := removeServerNamePort(serverAuthority)
 	timeoutCtx, cancel := context.WithTimeout(ctx, GetS2ATimeout())
 	defer cancel()
-	s2AStream, err := createStream(timeoutCtx, c.s2av2Address, c.getS2AStream)
+	var err error
+	var s2AStream stream.S2AStream
+	retryer := NewRetryer()
+	for {
```
matthewstevenson88, line 138: We have this same codeblock a lot. WDYT of having helper function that does this instead? It could accept like `func()` to keep it generic.

matthewstevenson88, line 138: Should we add an upper bound (e.g. 100) on the number of times we can enter this for-loop, to act as a safeguard against a bug in the retry library?

Similarly throughout.

internal/v2/s2av2.go
```diff
@@ -131,7 +132,22 @@ func (c *s2av2TransportCreds) ClientHandshake(ctx context.Context, serverAuthori
...
+		if err = ctx.Err(); err != nil {
+			break
+		}
+		s2AStream, err = createStream(timeoutCtx, c.s2av2Address, c.getS2AStream)
+		if backoff, shouldRetry := retryer.Retry(err); shouldRetry {
+			if sleepErr := retry.Sleep(ctx, backoff); sleepErr != nil {
```
matthewstevenson88, line 144: Should we log (maybe VLOG(1) if we're retrying)?

Similarly throughout.

internal/v2/s2av2.go
```diff
@@ -131,7 +132,22 @@ func (c *s2av2TransportCreds) ClientHandshake(ctx context.Context, serverAuthori
 	serverName := removeServerNamePort(serverAuthority)
 	timeoutCtx, cancel := context.WithTimeout(ctx, GetS2ATimeout())
 	defer cancel()
-	s2AStream, err := createStream(timeoutCtx, c.s2av2Address, c.getS2AStream)
+	var err error
+	var s2AStream stream.S2AStream
+	retryer := NewRetryer()
+	for {
+		if err = ctx.Err(); err != nil {
+			break
```
matthewstevenson88, line 140: Should we log whenever we break?

internal/v2/s2av2_e2e_test.go
```diff
@@ -335,6 +348,63 @@ func TestGRPCFallbackEndToEndUsingFakeS2AOverTCP(t *testing.T) {
...
+				Hostname: "test_rsa_server_identity",
+			},
+		},
+	}
+	serverAddr := startServer(t, serverS2AAddr, localIdentities)
+	fallbackServerAddr := startFallbackServer(t)
+	t.Logf("server running at address: %v", serverAddr)
+	t.Logf("fallback server running at address: %v", fallbackServerAddr)
+
+	// Finally, start up the client.
```
matthewstevenson88, line 384: To make sure I understand: in this test, we retry 5 times and then fall back. Is that correct?

Is it also possible to add an e2e test where we e.g. retry once and then succeed?

Similarly for the other e2e test.

s2a.go
matthewstevenson88: For my own understanding, why are we using 2 different contexts? Namely, `ctx` and `timeoutCtx`.

xmenxk: should use the the same context, thanks for catching this.

internal/v2/s2av2.go
xmenxk: updated added a helper function in retry.go
PTAL
`func Run(ctx context.Context, f func() error) {`

xmenxk: Thanks, updated
the retryer's internal limit is 5, I added a limit of 10 to the for loop as you suggested.

internal/v2/s2av2.go
xmenxk: done

internal/v2/s2av2.go
xmenxk: done

internal/v2/s2av2_e2e_test.go
xmenxk: that's right.

it's a little difficult to test the scenario of retry once and then succeed, with our fake s2a server.
Instead I was able to add a test for this logic in retry_test.go, let me know if that's okay:)

matthewstevenson88 (approved):
LGTM. Thanks for the PR Kui!

PR #117, opened 2023-08-22
s2a-go can talk to s2a-grpc-server in mtls

xmenxk:
s2a-go client can talk to s2a grpc server that requires mtls
This is done by adding an extra field `TransportCreds credentials.TransportCredentials` to s2a_options.
When `TransportCreds` is set, s2a-go will use it when connecting to the S2A-gRPC server.


internal/handshaker/service/service_test.go
```diff
@@ -19,6 +19,7 @@
 package service
 
 import (
+	"context"
```
matthewstevenson88, line 22: Can we add a unit test where the transport credentials are non-nil?

internal/record/ticketsender.go
```diff
@@ -83,7 +83,7 @@ func (t *ticketSender) sendTicketsToS2A(sessionTickets [][]byte, callComplete ch
 					t.ensureProcessSessionTickets.Done()
 				}
 			}()
-			hsConn, err := service.Dial(t.hsAddr)
+			hsConn, err := service.Dial(context.Background(), t.hsAddr, nil)
```
matthewstevenson88, line 86: Please add a comment explaining why we're not doing anything interesting with the context or transport credentials here.

internal/v2/s2av2.go
```diff
@@ -99,7 +101,7 @@ func NewClientCreds(s2av2Address string, localIdentity *commonpbv1.Identity, ver
 
 // NewServerCreds returns a server-side transport credentials object that uses
 // the S2Av2 to establish a secure connection with a client.
-func NewServerCreds(s2av2Address string, localIdentities []*commonpbv1.Identity, verificationMode s2av2pb.ValidatePeerCertificateChainReq_VerificationMode, getS2AStream func(ctx context.Context, s2av2Address string) (stream.S2AStream, error)) (credentials.TransportCredentials, error) {
+func NewServerCreds(s2av2Address string, s2aCreds credentials.TransportCredentials, localIdentities []*commonpbv1.Identity, verificationMode s2av2pb.ValidatePeerCertificateChainReq_VerificationMode, getS2AStream func(ctx context.Context, s2av2Address string) (stream.S2AStream, error)) (credentials.TransportCredentials, error) {
```
matthewstevenson88, line 104: nit: Can we rename this parameter to "transportCreds"?

Rationale: "s2aCreds" is what we're building as the output of this function.

Similarly above and below.

matthewstevenson88 (commented):
Thanks for the PR Kui! Generally looks good, just a few minor things.

internal/handshaker/service/service_test.go
xmenxk: done

internal/record/ticketsender.go
xmenxk: done!

internal/v2/s2av2.go
xmenxk: thanks , updated the naming accordingly.

matthewstevenson88 (approved):

PR #119, opened 2023-09-01
remove appengine dialer

xmenxk:
remove unused appengine dialer code and its dependency


go.mod
```diff
@@ -3,12 +3,12 @@ module github.com/google/s2a-go
...
-	golang.org/x/text v0.3.8 // indirect
-	google.golang.org/appengine v1.4.0
-	google.golang.org/grpc v1.45.0
-	google.golang.org/protobuf v1.26.0
+	github.com/golang/protobuf v1.5.3
+	github.com/google/go-cmp v0.5.9
+	golang.org/x/crypto v0.12.0
+	golang.org/x/net v0.14.0 // indirect
+	golang.org/x/sync v0.2.0
+	google.golang.org/genproto/googleapis/rpc v0.0.0-20230822172742-b8732ec3820d // indirect
```
matthewstevenson88, line 11: Why did this line get added when we're removing the appengine dependency?

I think this is the source of the giant go.sum diffs.

xmenxk: I also ran `go get -u` to update the modules

The -u flag instructs get to update modules providing dependencies of packages named on the command line to use newer minor or patch releases when available.

matthewstevenson88 (approved):

go.mod
matthewstevenson88: SGTM, just wanted to check. :)

PR #138, opened 2024-05-24
Clean up protos

rmehta19:
Summary of changes:
- add identity to v2 common.proto
- tokenmanager accepts v1 and v2 identity
- regenerate all protos [1]
- in s2a.go, use proto.Marshal / Unmarshal to convert v1 identity to v2 identity
- remove usage of v1 common.proto from v2 code

[1]: example commands to regenerate protos (for reference):
```
protoc -I . --proto_path=internal/proto/v2/s2a --go_out=internal/proto/v2/s2a_go_proto internal/proto/v2/s2a/s2a.proto
```

```
protoc --go-grpc_out=internal/proto/v2/s2a_go_proto internal/proto/v2/s2a/s2a.proto
```


internal/tokenmanager/tokenmanager.go
```diff
@@ -65,6 +66,14 @@ func (m *singleTokenAccessTokenManager) DefaultToken() (string, error) {
 }
 
 // Token always returns the token managed by the singleTokenAccessTokenManager.
-func (m *singleTokenAccessTokenManager) Token(*commonpb.Identity) (string, error) {
+func (m *singleTokenAccessTokenManager) Token(identity interface{}) (string, error) {
```
rmehta19, line 69: Some notes:
- can't use generics here: https://go.googlesource.com/proposal/+/refs/heads/master/design/43651-type-parameters.md#no-parameterized-methods
- can't use overloading: https://go.dev/doc/faq#overloading

Other option: have two methods: ```Token(*commonpbv1.Identity)``` and ```TokenV2(*commonpb.Identity)```

internal/proto/v2/common/common.proto
```diff
@@ -53,3 +53,25 @@ enum AlpnProtocol {
   ALPN_PROTOCOL_HTTP2 = 2;
   ALPN_PROTOCOL_HTTP1_1 = 3;
 }
+
+message Identity {
```
matthewstevenson88, line 57: It looks like we have duplicate copies of the Identity message - is that intentional?

xmenxk:
> Summary of changes:
> 
> * add identity to v2 common.proto
> * tokenmanager accepts v1 and v2 identity
> * regenerate all protos [1]
> * in s2a.go, use proto.Marshal / Unmarshal to convert v1 identity to v2 identity
> * remove usage of v1 common.proto from v2 code
> 
> [1]: example commands to regenerate protos (for reference):
> 
> ```
> protoc -I . --proto_path=internal/proto/v2/s2a --go_out=internal/proto/v2/s2a_go_proto internal/proto/v2/s2a/s2a.proto
> ```
> 
> ```
> protoc --go-grpc_out=internal/proto/v2/s2a_go_proto internal/proto/v2/s2a/s2a.proto
> ```

We have a script for proto generation at:
 https://github.com/google/s2a-go/blob/main/tools/proto/regenerate_proto.sh
 
 Can we use that script and apply any updates if needed? thanks

internal/proto/v2/common/common.proto
xmenxk: I wonder if it makes sense to `move` the Identity here, instead of creating a new copy of definition?

rmehta19: Yes -- duplicating ```Identity``` messsage was intentional in ```common.proto``` and ```v2/common.proto```, as discussed offline a few month's ago:
 "we could also add the Identity message to the V2 files...combining the two common.proto files into one is probably infeasible, I think, because the enums are incompatible". This change aligns the v2 protos with the copy in grpc-proto being used by the S2A Java client: https://github.com/grpc/grpc-proto/pull/149

Re: I wonder if it makes sense to move the Identity here, instead of creating a new copy of definition?
> I think this would be possible. We could create ```identity.proto``` which just contains the ```Identity``` message, and this could be used by both v1 and v2. IIRC, the goal as to not share protos between v1 and v2, but if we are ok with ```identity.proto``` being shared between the two, I don't see a problem.

rmehta19:
> > Summary of changes:
> > 
> > * add identity to v2 common.proto
> > * tokenmanager accepts v1 and v2 identity
> > * regenerate all protos [1]
> > * in s2a.go, use proto.Marshal / Unmarshal to convert v1 identity to v2 identity
> > * remove usage of v1 common.proto from v2 code
> > 
> > [1]: example commands to regenerate protos (for reference):
> > ```
> > protoc -I . --proto_path=internal/proto/v2/s2a --go_out=internal/proto/v2/s2a_go_proto internal/proto/v2/s2a/s2a.proto
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```
> > protoc --go-grpc_out=internal/proto/v2/s2a_go_proto internal/proto/v2/s2a/s2a.proto
> > ```
> 
> We have a script for proto generation at: https://github.com/google/s2a-go/blob/main/tools/proto/regenerate_proto.sh
> 
> Can we use that script and apply any updates if needed? thanks

Done -- thanks for pointing this out! Latest commit contains result of running script. The only changes are to the example protos.

internal/proto/v2/common/common.proto
matthewstevenson88: Thanks! I had forgotten the details of this discussion. :)

matthewstevenson88 (approved):

s2a.go
```diff
@@ -134,20 +144,38 @@ func NewServerCreds(opts *ServerOptions) (credentials.TransportCredentials, erro
...
+	var v2LocalIdentities []*commonpb.Identity
+	for _, marshalledLocalIdentity := range marshalledLocalIdentities {
+		v2LocalIdentity := &commonpb.Identity{}
+		err := proto.Unmarshal(marshalledLocalIdentity, v2LocalIdentity)
+		if err != nil {
+			return nil, err
+		}
+		v2LocalIdentities = append(v2LocalIdentities, v2LocalIdentity)
+	}
+
```
xmenxk, line 177: Do you think it would be better to have a separate toProtoIdentity() function for the v2 identity? instead of doing the marshal & un-marshal conversion.

My understanding is that at some point in the future, the v1/legacy code can be removed, and the marshal/unmarshal logic will need to be replaced anyway. wdyt?

internal/record/ticketsender_test.go
```diff
@@ -85,8 +86,16 @@ func (m *fakeAccessTokenManager) DefaultToken() (string, error) {
...
+func (m *fakeAccessTokenManager) Token(identity interface{}) (string, error) {
+	switch v := identity.(type) {
+	case *commonpbv1.Identity:
+		// valid type.
+	case *commonpb.Identity:
+		// valid type.
+	default:
+		return "", fmt.Errorf("Incorrect identity type: %v", v)
+	}
+	if identity == nil || cmp.Equal(identity, &commonpbv1.Identity{}, protocmp.Transform()) {
```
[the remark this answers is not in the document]
rmehta19: Thanks for catching this -- done!

s2a.go
rmehta19: My understanding is the same as yours. I have added a new ```toV2ProtoIdentity``` function in s2a_options.go, and replaced the marshal/unmarshal logic with this.

xmenxk (approved):

PR #144, opened 2024-07-02
Change proto tags when renaming Identity field

rmehta19:


matthewstevenson88:
Thanks for the PR @rmehta19! Please address the test failures, but otherwise LGTM.

rmehta19:
Thanks Matt! Fixed the failures.
Source: I ran https://github.com/google/s2a-go/blob/main/tools/proto/regenerate_proto.sh as this PR updates protos. It looks like running this scripts causes imports in gen proto files to be

```
common_go_proto "github.com/google/s2a/internal/proto/common_go_proto"
```

instead of:

```
common_go_proto "github.com/google/s2a-go/internal/proto/common_go_proto"
```

To fix, I ran the script to regenerate the proto files, and then went through each modified generated proto file and manually changed it to to the latter.

matthewstevenson88 (approved):

PR #151, opened 2024-12-05
Add extra parameters allowing user to configure the stream to S2A

zeromath:
This allows us to pass necessary parameters.

### DIFF
--- a/stream/s2a_stream.go
+++ b/stream/s2a_stream.go
@@ -20,6 +20,8 @@
 package stream
 
 import (
+	"context"
+
 	s2av2pb "github.com/google/s2a-go/internal/proto/v2/s2a_go_proto"
 )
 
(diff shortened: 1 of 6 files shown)


stream/s2a_stream.go
```diff
@@ -32,3 +34,6 @@ type S2AStream interface {
 	// Closes the channel to the S2A server.
 	CloseSend() error
 }
+
+// GetS2AStream type is for generating an S2AStream interface for talking to the S2A server.
+type GetS2AStream func(ctx context.Context, s2av2Address string, opts ...string) (S2AStream, error)
```
xmenxk, line 39: do we want to define some `Option` struct here or string tyoe is sufficient?
`opts ...string)`

zeromath: Yeah, I think this allows the greatest flexibility.

If we add an `Option` struct instead, every time if we want to add a new field to this Option we need to create a PR. Instead now we can pass as many parameters as we want and rely on the actual implementation to interpret these parameters.

xmenxk (approved):

PR #155, opened 2025-04-08
HTTP Client TLS configs should send identity to the S2A server

gscert:


gscert:
Hey @zeromath, can you PTAL?

zeromath:
Shall we rename this to saying that this change is for HTTP client?

Other than this, everything else looks good to me.

s2a.go
```diff
@@ -320,6 +320,10 @@ func NewTLSClientConfigFactory(opts *ClientOptions) (TLSClientConfigFactory, err
 	if opts.EnableLegacyMode {
 		return nil, fmt.Errorf("NewTLSClientConfigFactory only supports S2Av2")
 	}
+	localIdentity, err := toV2ProtoIdentity(opts.LocalIdentity)
+	if err != nil {
```
xmenxk, line 324: so if a LocalIdentity is not specified, then we get an error?

gscert: I believe toV2ProtoIdentity only returns an error if an unsupported identity is provided. I added some tests to make sure.

xmenxk (approved):

PR #159, opened 2026-09-18
Allow clients to select ALPN protocols instead of forcing HTTP/2

sruthi-talluri:
GetTLSConfigurationForClient has hardcoded `NextProtos to {"h2"}`, which was added before when s2a-go was only used for gRPC. The HTTP entry points added later, `NewTLSClientConfigFactory` and `NewS2ADialTLSContextFunc`, inherited that default, so an HTTP/1.1 client gets a connection that has negotiated h2 via ALPN. net/http then writes HTTP/1.1 onto it and the first response fails to parse:

```
  net/http: HTTP/1.x transport connection broken: malformed HTTP response
  "\x00\x00$\x04\x00\x00\x00\x00\x00\x00\x05\x00\x10..."
  ```

which is an HTTP/2 SETTINGS frame. There is currently no way to avoid this other than rebuilding the tls.Config by hand after Build returns.

This adds `ClientOptions.NextProtos`, applied whenever a client `tls.Config` is built, and `TLSClientConfigOptions.NextProtos` as a per-Build override. `NewS2ADialTLSContextFunc` goes through `NewTLSClientConfigFactory`, so it picks the option.

Behavior is unchanged when neither is `set: nextProtosOrDefault` falls back to `{"h2"}`. The gRPC ClientHandshake path in `internal/v2/s2av2.go` passes `nil` explicitly, since that transport requires HTTP/2 and must not be overridable. The returned slice is copied so a later mutation by the caller cannot alter a live `tls.Config`.

### CODE REVIEW

File: internal/v2/tlsconfigstore/tlsconfigstore.go [hunk 1]
```diff
@@ -42,8 +42,21 @@ const (
 	h2 = "h2"
 )
 
+// nextProtosOrDefault returns the ALPN protocols to offer during the
+// handshake. S2A has historically offered HTTP/2 only, which is required by
+// the gRPC transport, so that remains the default when no protocols are
+// requested.
```
rmehta19, line 48: nit: We can remove this second sentence

internal/v2/tlsconfigstore/tlsconfigstore.go
```diff
@@ -42,8 +42,21 @@ const (
...
+func nextProtosOrDefault(nextProtos []string) []string {
+	if len(nextProtos) == 0 {
+		return []string{h2}
+	}
+	return append([]string(nil), nextProtos...)
```
rmehta19, line 53: Can this be replaced with `slices.Clone(nextProtos)`?

s2a.go
```diff
@@ -305,6 +305,11 @@ type TLSClientConfigOptions struct {
 	//			ServerName: "example.com",
 	//		})
 	ServerName string
+
+	// NextProtos is the list of ALPN protocols offered during the TLS
+	// handshake. If empty, ClientOptions.NextProtos is used; if that is also
+	// empty, HTTP/2 ("h2") is offered.
+	NextProtos []string
```
rmehta19, line 312: NextProtos param is added in TLSClientConfigOptions, s2aTLSClientConfigFactory and ClientOptions (in s2a_options.go).

 Can it just be in TLSClientConfigOptions?

s2a.go
```diff
@@ -359,6 +366,7 @@ type s2aTLSClientConfigFactory struct {
 	getStream                 stream.GetS2AStream
 	// localIdentity should only be used by the client.
 	localIdentity *commonpb.Identity
+	nextProtos    []string
 }
 
 func (f *s2aTLSClientConfigFactory) Build(
```
rmehta19, line 372: Can we update s2a_test.go and s2a_e2e_test.go?

bargielavictor82-commits:
***@***.***

internal/v2/tlsconfigstore/tlsconfigstore.go
anushka567: Addressed in https://github.com/google/s2a-go/pull/161

internal/v2/tlsconfigstore/tlsconfigstore.go
anushka567: Addressed in https://github.com/google/s2a-go/pull/161

s2a.go
anushka567: Addressed in https://github.com/google/s2a-go/pull/161

s2a.go
anushka567: We need `NextProtos` in `ClientOptions` because `NewS2ADialTLSContextFunc` takes `*ClientOptions` and invokes `factory.Build(ctx, &TLSClientConfigOptions{ServerName: serverName})` internally on every dial in the background.

Callers configuring `http.Transport` via `NewS2ADialTLSContextFunc` (such as GCSFuse) do not call `Build()` directly, so without `ClientOptions.NextProtos`, dialer users would have no way to configure ALPN. 

Having `NextProtos` in both places gives us the best of both worlds:
1. **`ClientOptions.NextProtos`**: Sets the default for the factory and makes `NewS2ADialTLSContextFunc` work out of the box for HTTP/1.1 clients.
2. **`TLSClientConfigOptions.NextProtos`**: Provides an optional per-`Build()` override when interacting with the low-level factory directly.



Following up in #161

PR #161, opened 2026-09-22
Allow clients to select ALPN protocols instead of forcing HTTP/2

anushka567:
# Description

This PR carries forward the changes from #159 (by @sruthi-talluri) to allow clients to select ALPN protocols instead of forcing HTTP/2, and addresses the review feedback on that PR.

## Background

`GetTLSConfigurationForClient` has hardcoded `NextProtos` to `{"h2"}` since #72, which was added when `s2a-go` was only used for gRPC. The HTTP entry points added later (`NewTLSClientConfigFactory` and `NewS2ADialTLSContextFunc`) inherited that default, so an HTTP/1.1 client gets a connection that has negotiated `h2` via ALPN. `net/http` then writes HTTP/1.1 onto it and the first response fails to parse:

```text
net/http: HTTP/1.x transport connection broken: malformed HTTP response
"\x00\x00$\x04\x00\x00\x00\x00\x00\x00\x05\x00\x10..."
```
which is an HTTP/2 `SETTINGS` frame.

## Changes

1. **ALPN Selection**:
   - Adds `ClientOptions.NextProtos` (applied when a client `tls.Config` is built via `NewS2ADialTLSContextFunc` and `NewTLSClientConfigFactory`).
   - Adds `TLSClientConfigOptions.NextProtos` as a per-`Build()` override.
   - Falls back to `{"h2"}` when unset, and strictly enforces HTTP/2 for gRPC client handshakes.

2. **Addressed Review Feedback from #159**:
   - Replaced manual slice append with `slices.Clone(nextProtos)` in `internal/v2/tlsconfigstore/tlsconfigstore.go`.
   - Cleaned up doc comment on `nextProtosOrDefault`.
   - Added unit tests for `NewTLSClientConfigFactory` with custom `NextProtos` in `s2a_test.go`.
   - Added end-to-end HTTP/1.1 negotiation test in `s2a_e2e_test.go`.

Co-authored-by: Sruthi Talluri <sruthitalluri@google.com>


internal/v2/s2av2_e2e_test.go
```diff
@@ -513,7 +513,7 @@ func TestNewClientTlsConfigWithCustomS2AStream(t *testing.T) {
 		return s2ATestStream{debug: "test S2A stream"}, nil
 	}
 
-	_, err := NewClientTLSConfig(ctx, s2aAddr, nil, tokenManager, s2av2pb.ValidatePeerCertificateChainReq_CONNECT_TO_GOOGLE, "test_server_name", nil, getS2AStream, identity)
+	_, err := NewClientTLSConfig(ctx, s2aAddr, nil, tokenManager, s2av2pb.ValidatePeerCertificateChainReq_CONNECT_TO_GOOGLE, "test_server_name", nil, getS2AStream, identity, nil)
 	if err != nil {
 		t.Errorf("NewClientTLSConfig() failed: %v", err)
 	}
```
rmehta19: Can we add a test here where nextProtos is non-nil.

s2a.go
```diff
@@ -367,7 +375,11 @@ func (f *s2aTLSClientConfigFactory) Build(
 	if opts != nil && opts.ServerName != "" {
 		serverName = opts.ServerName
 	}
-	return v2.NewClientTLSConfig(ctx, f.s2av2Address, f.transportCreds, f.tokenManager, f.verificationMode, serverName, f.serverAuthorizationPolicy, f.getStream, f.localIdentity)
+	nextProtos := f.nextProtos
+	if opts != nil && len(opts.NextProtos) > 0 {
+		nextProtos = opts.NextProtos
+	}
```
rmehta19: nit: can combine with opts.ServerName check above:

`
if opts != nil {
        if opts.ServerName != "" {
		 serverName = opts.ServerName
        }
        // check next protos
}
`

rmehta19:
Can we manually E2E test this? We have an example folder in the top level directory which may contain some useful components for testing

matthewstevenson88 (approved):

rmehta19 (approved):

